PR #32, opened 2020-10-26
Revised JSON schema with examples

rd2:


rd2:
This is a first step in the JSON I/O work. The examples are [valid](https://jsonschemalint.com/#!/version/draft-04/markup/json) against the schema - more work is needed to validate the links between OpenStudio & TBD objects. More to come ...

PR #33, opened 2020-10-30
JSON Schema validation examples

rd2:
Fine-tuning of draft TBD JSON Schema, with [valid](https://jsonschemalint.com/#!/version/draft-04/markup/json) examples. More to come (I suspect) when deploying post-schema validation (e.g. duplicate keys, case (in)sensitivity).


lib/measures/tbd/resources/psi.rb
```diff
@@ -10,15 +11,43 @@
...
+
+    # The following are defaults (* stated). Users may edit these defaults,
+    # add new KHI pairs, or even read-in other KHI pairs on file.
+    # Units are in W/K.
+    @point[ "poor (BC Hydro)" ]         = 0.900.freeze # detail 5.7.2 BETB
+    @point[ "regular (BC Hydro)" ]      = 0.500.freeze # detail 5.7.4 BETB
+    @point[ "efficient (BC Hydro)" ]    = 0.150.freeze # detail 5.7.3 BETB
+    @point[ "code (Quebec)" ]           = 0.500.freeze # art. 3.3.1.3. NECB-QC
+    @point[ "(non thermal bridging)" ]  = 0.000.freeze
+  end
```
macumber, line 38: If you are trying to make sure that people can't modify the hash I think you want `@point.freeze` after you set the defaults.  `0.900.freeze` just means you can't change that literal value which you couldn't anyway.

json/tbd_PSI_combo.json
```diff
@@ -25,16 +25,15 @@
       "grade": 0.2
     }
   ],
-  "units": [
+  "unit": [
     {
-      "id": "building",
-      "psis": "Awesome"
+      "psi": "Awesome"
```
macumber, line 30: Is this supposed to be a failing test?

spec/tbd_tests_spec.rb
```diff
@@ -1566,3 +1566,414 @@
...
+          "description": "Surface(s) edge as thermal bridge",
+          "type": "object",
+          "properties": {
+            "psi": {
+              "title": "PSI-set identifier",
+              "type": "string"
+            },
+            "type": {
+              "title": "PSI-set type, e.g. 'parapet'",
+              "type": "string"
```
macumber, line 1729: Do you want to enumerate the allowable types?

spec/tbd_tests_spec.rb
```diff
@@ -1566,3 +1566,414 @@
...
+            },
+            "length": {
+              "title": "Edge length (m), 10cm min",
+              "type": "number",
+              "minimum": 0.1
```
macumber, line 1734: I'd make minimum be 0, you can ignore them later but you will have edges that are small

spec/tbd_tests_spec.rb
```diff
@@ -1566,3 +1566,414 @@
...
+        },
+        "KHI": {
+          "description": "KHI-value (in W/K) for a point thermal bridge",
+          "type": "object",
+          "properties": {
+            "id": {
+              "title": "Unique KHI identifier",
+              "type": "string"
+            },
+            "point": {
```
macumber, line 1707: Would you want to enumerate the types of KHI beyond just point?  The types of PSI correspond to real world things.

spec/tbd_tests_spec.rb
```diff
@@ -1566,3 +1566,414 @@
...
+          "additionalProperties": false,
+          "required": [
+            "psi",
+            "type",
+            "length",
+            "surfaces"
+          ]
+        },
+        "Surface": {
+          "description": "Surface default PSI-set (optional)",
```
macumber, line 1755: I'm confused about whether this Surface object refers to an OS/E+ surface or if it is like a DefaultPSIKHISet?

spec/tbd_tests_spec.rb
```diff
@@ -1566,3 +1566,414 @@
...
+              "items": {
+                "type": "string"
+              },
+              "minItems": 1,
+              "uniqueItems": true
+            }
+          },
+          "additionalProperties": false,
+          "required": [
+            "psi",
```
macumber, line 1748: If psi is required for each edge, how do you specify that an edge uses a default value?

spec/tbd_tests_spec.rb
```diff
@@ -1566,3 +1566,414 @@
...
+              "id": "support",
+              "count": 4
+            }
+          ],
+          "psi": "good"
+        }
+      ],
+      "unit": [
+        {
+          "psi": "compliant"
```
macumber, line 1915: What is this example showing?  The unit is compliant, the surface is good, and the edge is compliant?  Would the other edges just not be listed here and they would adopt the values from their Space or Unit?  How would we know the type of edge for an edge that isn't listed here?  Is there a default edge type?

spec/tbd_tests_spec.rb
```diff
@@ -1566,3 +1566,414 @@
...
+            "front wall"
+          ]
+        }
+      ],
+      "surfaces": [
+        {
+          "id": "front wall",
+          "khis": [
+            {
+              "id": "column",
```
macumber, line 1902: These khi values just apply to the surface, they don't have to be associated with a physical point?

json/tbd_full_PSI.json
```diff
@@ -11,6 +11,7 @@
       "convex": 0.700,
       "balcony": 1.000,
       "party": 0.700,
+      "party": 0.800,
```
macumber, line 14: This repeated key "party" should not be allowed right?

spec/tbd_tests_spec.rb
macumber: https://json-schema.org/understanding-json-schema/reference/generic.html#id4

spec/tbd_tests_spec.rb
macumber: Do you just not list edges that are defaulted?

lib/measures/tbd/resources/psi.rb
rd2: Good call. I was doing it right for the PSI sets, but not here. Will correct.

spec/tbd_tests_spec.rb
rd2: Good point. I'll use the [enum](https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values) keyword. I was planning to add a post-schema validation control to flag unrecognized types here, i.e. "doorjamb" - enum is far better. Thanks.

spec/tbd_tests_spec.rb
rd2: The Topolys tolerances should filter out any edge under 1 cm (and we've flirted with moving this up to an inch). But you're right, as Topolys can/will generate 2-inch or 3-inch edges depending on OSM geometry. It's bad design on the user's part (or any third-party app), but that's irrelevant. Agree to drop to "0", but will likely add an "exclusiveMinimum" property - we should never come across a strict "0" value.

spec/tbd_tests_spec.rb
rd2: That's a good question. In short "no", but I think it's best to state why as we may want to revisit this. So true, enumerated PSI types do correspond to real world things. But more importantly, they directly relate to keywords in current standards and codes, which often reserve particular treatment for each thermal bridge (TB) type. So the enumerated PSI types are key to (automatically) generating _valid_ reference/budget buildings. For instance, pre-2017 Canadian energy codes offer exceptions to slab-wall edges ("rimjoist"), while the QC energy code has different requirements for slab-wall ("rimjoist"), roof-wall ("parapet"), and slab projections ("balcony"). The "fenestration" type could one day (?) be expanded (as an option) to differentiate between "jamb", "sill" and "lintel" - but Industry isn't there yet, so there's little value in pre-emptively expanding TB types IMO. For _point_ TBs, their current treatment is quite coarse - I don't see any differentiation in codes/standards between a series of columns punching through an exposed floor vs roof supports for HVAC equipment. For instance, the QC energy code has a default 0.5 W/K requirement for any major _point_ TB, regardless of application. It's better than nothing, but I expect more granularity in the future. Similarly to fenestration vs jamb-sill-lintel, once someone can point me to a code or standard that differentiates between point TB types, I'll gladly add that in. In the meantime, TBD doesn't hold enumerated KHI types.

spec/tbd_tests_spec.rb
rd2: Maybe this [example](https://github.com/rd2/tbd/blob/json/json/tbd_PSI_combo2.json) helps. For starters (or maybe more of a reminder), "psis" (PSI definition) and "khis" (KHI definition) are libraries, analogous to default construction sets in OpenStudio. If one is not interested in the few default sets provided with TBD (e.g. BC Hydro sets), one can use the JSON file (as input) to specify their own default "Unit" PSI-set, which in turn would be used by TBD to initially set all thermal bridges (TBs) in a building (based on enumerated allowable PSI types, e.g. "parapet" vs "fenestration"). One can also specify default Story, Space and/or Surface PSI-sets, overriding the default Unit PSI-set. So for instance, any Surface set will supersede its immediate Space set (if present). An Edge PSI-set is slightly different in that it allows users to hijack a specific edge they suspect will be (or notice is) misdiagnosed by TBD (e.g. "concave" corner rather than a "party" wall edge - it'll happen), or even add an edge that isn't even in the OSM geometry (similar to adding a _point_ TB). So in the cited example, all TBs are by default initially processed using the building Unit PSI-set ("Awesome"), yet the "ok" set shall instead be used for the "ground-floor restaurant South-wall" (overriding the "Awesome set"). This is very similar to default constructions vs surface-specified constructions. Finally, any edge connecting the "ground-floor restaurant West-wall" and the "ground-floor restaurant party wall", and 4.3m in length, will be processed using the "Part wall edge", also overriding the Unit set. If no such edge(s) exist(s) in Topolys/TBD model, then TBD will process it as an additional edge (similarly to a _point_ TB).

spec/tbd_tests_spec.rb
rd2: In 99% of cases, one shouldn't see "edges" in a TBD JSON file (when used as input ... for output, that's a whole other thing).

The "edges" input option is offered for (i) the odd cases where a misdiagnosis occurs (i.e. TBD misidentifies the edge type) or (ii) when an edge isn't in the OSM file. These are fairly rare cases.

spec/tbd_tests_spec.rb
rd2: So for starters, there's a mistake in the example. The first entry for "psis" should read :

```
{
  "id": "good",
  "parapet": 0.5,
  "party": 0.9
},
```

... my bad. Will correct in next commit. So the way one should read this input TBD JSON file:

- By default, all Topolys-detected edges shall be processed by TBD using the building unit default set "compliant".
- In addition, the surface "front wall" shall inherit 3x columns and 4x supports as _point_ thermal bridges (TBs). Wall with columns? [Flying buttresses](https://en.wikipedia.org/wiki/Flying_buttress)!
- Furthermore, all parapets and party wall edges of that same "front wall" shall not default to the "compliant" PSI-set, but rather the "good" set ... all other edge types (e.g. "fenestration") of the "front wall" shall continue to rely on the default "compliant" set.
- Exceptionally, a single party wall edge (2.5 m in height) is treated particularly efficiently, more evocative of the "compliant" PSI-set. So that specific edge shall not be processed by TBD using the "front wall"-specific "good" set, but revert back to the "compliant" set.

Is this a somewhat extreme case of customization? Yes! But something similar will undoubtedly be requested.

spec/tbd_tests_spec.rb
rd2: Correct. Their XYZ location along the "front wall" isn't relevant.

json/tbd_full_PSI.json
rd2: Correct. [Here](https://github.com/rd2/tbd/blob/1f88452b05f6e5fc4654f4256e306a4fc4681a60/spec/tbd_tests_spec.rb#L1952), I left (myself) a reminder to address default JSON Schema validation with regards to repeated keys. Post-schema validation remains unavoidable.

rd2:
@macumber : Thanks for the detailed review. It may take a few other iterations to nail this.

spec/tbd_tests_spec.rb
rd2: Although my initial reply correctly describes the context where one would expect to encounter custom "edges" in a TBD JSON file (as input), I just noticed I wasn't directly answering your 2 questions. Here goes:

**1.** In circumstances where a TBD JSON file is used as input, there are **2** cases (**a** & **b**) to consider for "edges":

**a.** edges that are **absent** from the OSM geometry (and hence absent from the Topolys model): Impossible for TBD to diagnose the thermal bridge (TB) type (e.g. "rimjoist" ? "balcony" ?) on its own - users must inform TBD of the TB type (e.g. "balcony"). That being said, I see your point in this specific instance: users should nonetheless have the option of _defaulting_ the edge PSI-value to whatever "Surface", "Space", "Story" or building "Unit" default set would apply. I have since removed the "psi" key from the _required_ list of _properties_ for "Edges". (Good call, thanks). Upcoming commit ...

**b.** edges, although **present** in the Topolys model, that users nonetheless want to (i) override the TBD-identified TB type (e.g. "balcony" rather than "rimjoist"), or (ii) override the assigned PSI-value (superseding whatever default PSI-set would apply). Situation (i) is similar to case (a) above: users should also have the option of _defaulting_ the PSI-value. However in situation (ii), one would expect users to specify the PSI-value/set ... as this is the likely objective.

**2a.** "Do you just not list edges that are defaulted?" ... Based on the previous answer, one can now list any edge (present or absent from the Topolys model), while _defaulting_ (or not) its PSI-value.

**2b.** "How would you know the type of the defaulted edges?" ... in cases 1a & 1b(i) above, this is likely what a user would actually want, i.e. to override an edge's TBD-assigned type (e.g. "balcony" instead of "rimjoist"). This is analogous to resetting an OpenStudio surface type (e.g. a slanted "floor", instead of an auto-assigned "wall"). For case 1b(ii), that's simply how one (re)sets the PSI-value of an edge: PSI-set + TB type. It would be odd if a user could pinpoint an individual edge to customize and not know it's nature ...

In light of this, I have also decided to remove "length" as a _required_ property. A list of 2x surfaces (that share the edge) should be enough to pinpoint the edge in the Topolys model. However, as more than one edge can be shared between 2x surfaces, a provided "length" would narrow down the search. For cases 1a, the "length" is absolutely required - so will be looking more at post-schema validation in that case. Thanks for raising these questions.

PR #34, opened 2020-11-07
JSON-input internal validation - part 1

rd2:


rd2:
This is Part 1 (of 2) on _internal_ validation of TBD JSON input. _Internal_ refers to logical consistency of file content (not flagged by schema validation per se), either concerning:

- TBD structure & built-in default values
- internal logic of the file itself (e.g. a surface's custom PSI-set requires a matching library PSI-set on file)
- correspondence with OSM objects (e.g. a surface's id requires a matching OSM surface name)

PR #35, opened 2020-11-08
JSON-input internal validation - part 2a

rd2:


rd2:
Post-schema validation of TBD JSON input file content ...

First subtask 2a: on internal logic, e.g. surface's custom PSI-set requires a matching library PSI-set on file.

Second (final) subtask on OSM correspondence, e.g. surface's id requires a matching OSM surface name ... soon!

PR #36, opened 2020-11-16
JSON cutomization - 50% tested

rd2:
Most of the customization functionality linked to TBD JSON (as input) is complete. User-defined PSI & KHI library additions are well tested. The custom `:unit` (project-wide PSI defaults) is also well tested and working. The custom `:story`, `:space`, `:surface` & `:edge` features are integrated yet only @50% tested. A grind ... a few hours more.

### DIFF
--- a/lib/measures/tbd/resources/psi.rb
+++ b/lib/measures/tbd/resources/psi.rb
@@ -17,7 +17,7 @@ end
 # www.bchydro.com/content/dam/BCHydro/customer-portal/documents/power-smart/
 # business/programs/BETB-Building-Envelope-Thermal-Bridging-Guide-v1-4.pdf
 
-# NECB-QC : Québec's energy code for new comemrcial buildings
+# NECB-QC : Québec's energy code for new commercial buildings
 # www2.publicationsduquebec.gouv.qc.ca/dynamicSearch/telecharge.php?type=1&file=72541.pdf
 
 class KHI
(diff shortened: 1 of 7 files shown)


lib/measures/tbd/resources/psi.rb
```diff
@@ -149,6 +149,173 @@ def append(p)
   end
 end
 
+def processTBDinputs(surfaces, edges, set, io_f, schema_f)
```
macumber, line 152: Is `io_f` an `IO` object or a file path?  It looks like a file path?  If so I'd suggest naming it path or something

lib/measures/tbd/resources/psi.rb
```diff
@@ -149,6 +149,173 @@ def append(p)
...
+      end
+
+    else
+      # Log severe warning: enable to parse (invalid) user TBD JSON file
+    end
+  else
+    # No (optional) user-defined TBD JSON input file.
+    # In such cases, "set" must refer to a valid PSI set
+    raise "processTBDinputs: PSI set #{set}?" unless psi.set.has_key?(set)
+    raise "Set #{set} rimjoist?"     unless psi.set[set].has_key?(:rimjoist)
```
macumber, line 305: Maybe make a method like `PSI::has_all_keys?`?

lib/measures/tbd/resources/psi.rb
rd2: Can do.

lib/measures/tbd/resources/psi.rb
rd2: Good idea.

PR #37, opened 2020-11-25
99% TBD JSON custom input validation + KHi's

rd2:
Validation of JSON customized inputs, e.g.:

- a space-specific PSi set superseding building PSi defaults
- a surface-specific PSi set superseding the previous space-specific PSi
- individual edges having custom types and/or PSi values, superseding surface-specific PSi set

Building story-specific PSI set override: future work (not much to it, just strapped for time).

Also added/tested point thermal bridges: KHi [in W/K] instead of PSI [in W/K per m] - done.

Next: exporting TBD/Topolys data via the TBD JSON file.


json/tbd_seb_n2.json
```diff
@@ -0,0 +1,47 @@
...
+  "khis": [
+    {
+      "id": "column",
+      "point": 0.5
+    },
+    {
+      "id": "support",
+      "point": 0.5
+    }
+  ],
```
rd2, line 31: "psis" & "khis" are libraries (referenced or not elsewhere on file).

json/tbd_seb_n2.json
```diff
@@ -0,0 +1,47 @@
...
+      "khis": [
+        {
+          "id": "column",
+          "count": 3
+        },
+        {
+          "id": "support",
+          "count": 4
+        }
+      ]
```
rd2, line 44: Point thermal bridges (in W/K) are surface-specific entries, relying on KHI library items.

json/tbd_seb_n3.json
```diff
@@ -0,0 +1,48 @@
...
+        },
+        {
+          "id": "support",
+          "count": 4
+        }
+      ],
+      "psi": "good"
+    }
+  ]
+}
```
rd2, line 48: Currently too many tests (initially useful for validating incremental variations while developing). Only the later versions (ex. n6, n7) will likely be kept at release.

lib/psi.rb
```diff
@@ -147,16 +147,33 @@ def append(p)
...
+    raise "'#{s}' rimjoist?"        unless @set[s].has_key?(:rimjoist)
+    raise "'#{s}' parapet?"         unless @set[s].has_key?(:parapet)
+    raise "'#{s}' fenestration?"    unless @set[s].has_key?(:fenestration)
+    raise "'#{s}' concave?"         unless @set[s].has_key?(:concave)
+    raise "'#{s}' convex?"          unless @set[s].has_key?(:convex)
+    raise "'#{s}' balcony?"         unless @set[s].has_key?(:balcony)
+    raise "'#{s}' party?"           unless @set[s].has_key?(:party)
+    raise "'#{s}' grade?"           unless @set[s].has_key?(:grade)
+  end
 end
```
rd2, line 162: Good suggestion.

lib/psi.rb
```diff
@@ -1080,10 +1059,10 @@ def processTBD(os_model, io_f, schema_f, psi_set)
   # completed/tested, so ignored for now ...
   # openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-2.9.0-doc/model/html/classopenstudio_1_1model_1_1_building_story.html
   if io
-    # if io.has_key?(:stories)    # ... will override :unit sets
+    # if io.has_key?(:stories)               # ... will override :unit sets
     # end
```
rd2, line 1064: Next step ...

lib/psi.rb
```diff
@@ -1092,24 +1071,32 @@ def processTBD(os_model, io_f, schema_f, psi_set)
         next unless io_p.set.has_key?(p)
 
         edges.values.each do |edge|
-          next unless edge.has_key?(:psi)
+          next unless edge.has_key?(:psi)   # open to transition edges TO DO ...
```
rd2, line 1074: By default, _mild_ transitions between 2 surfaces are ignored in TBD. In most cases such edges are either artefacts of the model (e.g. from tessallation) or subtle differences in construction (e.g. change in cladding, or an extra inch of insulation), and as such should not be considered (by default) as _major_ thermal bridges. There will be cases where this presumption is flat out wrong (e.g. structural transitional joints, curtain wall spandrel perimeters along other wall types). Adding another PSI type is likely the best way moving forward.

lib/psi.rb
```diff
@@ -1213,14 +1222,34 @@ def processTBD(os_model, io_f, schema_f, psi_set)
...
 
   # Assign thermal bridging heat loss [in W/K] to each deratable surface.
-  surfaces.values.each do |surface|
+  surfaces.each do |id, surface|
     next unless surface.has_key?(:edges)
     surface[:heatloss] = 0
     surface[:edges].values.each do |edge|
       surface[:heatloss] += edge[:psi] * edge[:length]
     end
   end
```
rd2, line 1231: Oops. Forgot to switch back to values; id is only used here sometimes for testing.

lib/psi.rb
```diff
@@ -1213,14 +1222,34 @@ def processTBD(os_model, io_f, schema_f, psi_set)
...
+        next unless k.has_key?(:id)
+        next unless k.has_key?(:count)
+        next unless io_k.point.has_key?(k[:id])
+        next unless io_k.point[k[:id]] > 0.000
+        surface[:heatloss] = 0 unless surface.has_key?(:heatloss)
+        surface[:heatloss] += io_k.point[k[:id]] * k[:count]
+      end
+    end
+  end
+
```
rd2, line 1252: The above code is where point thermal bridges are appended to surfaces.

json/tbd_seb_n2.json
macumber: So this is saying that the surface named "Entryway  Wall 5" is derated by two point bridges, a column and a support?

macumber: Is the assumption that these point bridges are always to the outside environment?  E.g. a column or support inside a conditioned zone don't count as thermal bridges?

json/tbd_seb_n3.json
macumber: Where would you reference the default `psi` set for the entire model?

json/tbd_seb_n4.json
```diff
@@ -0,0 +1,67 @@
...
+        {
+          "id": "column",
+          "count": 3
+        },
+        {
+          "id": "support",
+          "count": 4
+        }
+      ],
+      "psi": "(non thermal bridging)"
```
macumber, line 59: Is `(non thermal bridging)` a special keyword?  Could you just make a new `psi` set called `perfect` with no derating and reference that instead of using a keyword?

json/tbd_seb_n4.json
```diff
@@ -0,0 +1,67 @@
...
+      "id": "support",
+      "point": 0.0
+    }
+  ],
+  "edges": [
+    {
+      "psi": "bad",
+      "type": "fenestration",
+      "surfaces": [
+        "Entryway  Wall 5",
```
macumber, line 41: I am assuming that a `psi` specified for a specific edge overrides/replaces `psi` specified for the surface right?  It doesn't add/multiply with the surface `psi`?

lib/measures/tbd/resources/psi.rb
```diff
@@ -147,16 +147,33 @@ def append(p)
       end
     end
   end
+
+  def complete?(s)
```
macumber, line 151: Normally a `?` method would return true or false rather than raise an exception

json/tbd_seb_n3.json
macumber: If there is no default `psi` specified does that mean no derating?

lib/measures/tbd/resources/psi.rb
```diff
@@ -199,15 +216,7 @@ def processTBDinputs(surfaces, edges, set, io_f, schema_f)
...
-      raise "Unit #{p} mismatch?"     unless psi.set.has_key?(p)
-      raise "Unit #{p} rimjoist?"     unless psi.set[p].has_key?(:rimjoist)
-      raise "Unit #{p} parapet?"      unless psi.set[p].has_key?(:parapet)
-      raise "Unit #{p} fenestration?" unless psi.set[p].has_key?(:fenestration)
-      raise "Unit #{p} concave?"      unless psi.set[p].has_key?(:concave)
-      raise "Unit #{p} convex?"       unless psi.set[p].has_key?(:convex)
-      raise "Unit #{p} balcony?"      unless psi.set[p].has_key?(:balcony)
-      raise "Unit #{p} party?"        unless psi.set[p].has_key?(:party)
-      raise "Unit #{p} grade?"        unless psi.set[p].has_key?(:grade)
+      psi.complete?(p)
```
macumber, line 219: Here you would do `raise 'Not Complete' unless psi.complete?(p)`

json/tbd_seb_n4.json
```diff
@@ -0,0 +1,67 @@
...
+        "Sub Surface 8"
```
macumber, line 42: The approach of specifying edges by two surfaces (as opposed to an edge id) is growing on me, I like it because you can find that information from just the OSM/IDF.  Just to make sure, this `psi` would apply to all edges shared between these two surfaces in the case that they share more than one edge right?

json/tbd_seb_n3.json
macumber: Also, I don't think you can really ever have too many tests, I would recommend keeping them all.

json/tbd_seb_n2.json
rd2: 1. In the example, the surface "Entryway Wall 5" has 3x columns + 4x supports (e.g. signage). As both columns and supports represent 0.5 W/K, that means that all 7x point thermal bridges represent a total of 3.5 W/K.
2. Your 2nd question is an important one at this stage. Although Topolys handles all geometry, TBD will currently only process linear and point thermal bridges affecting outside-facing walls, roofs and exposed floors. Partitions separating spaces with design (and operating) temperatures that differ by e.g. more than 10°C should be insulated (by code), and most likely should include minor and major thermal bridging. I'm adding this to the enhancement list.

macumber: Cool, I missed the `count` property, that is a nice feature.

json/tbd_seb_n3.json
rd2: The (optional) :unit PSI set (on file) is the default one for the entire model. If it is absent from the JSON file, then the TBD Measure PSI set pull-down option is instead used as the default one for the entire model. If both are set/provided, the :unit one on file supersedes the pull-down option.

json/tbd_seb_n4.json
rd2: "(non thermal bridging)" is one of the 5 default PSI sets hardcoded in psi.rb, and so currently available in the TBD Measure PSI set pull-down options. With the input JSON file, anyone could add their own flavour, e.g. "perfect" with "0.000" W/K per m for each type.

json/tbd_seb_n4.json
rd2: Correct assumption. It supersedes the surface one, which supersedes the space one, etc. No addition. In the end, only one PSI type (e.g. :corner vs :fenestration ... not both) can end up characterizing an edge as-thermal-bridge.

lib/measures/tbd/resources/psi.rb
rd2: Noted. I simplify this accordingly, but maybe wait when I'll log mild/severe warnings on file.

lib/measures/tbd/resources/psi.rb
rd2: Got it.

json/tbd_seb_n4.json
rd2: You got it. The surface list is not a bad approach in the future for third-party GUIs (e.g. plugin), or even manual editing of the JSON file, e.g. a user can easily figure out linked surface names (e.g. OpenStudio plugin inspector, interacting with the geometry model under the OpenStudio Applications Geometry tab). To your 2nd question, correct - all edges matching the criteria (in the example above, all 4x :fenestration perimeter edges) will use the custom edge :psi and/or :type. To narrow down the targeted edge list, one can specify the edge length (e.g. "3.55", which would rule out the 2x shorter window jambs).

PR #40, opened 2020-11-26
Json - Minor changes in preparation for io hash JSON export.

rd2:


lib/measures/tbd/measure.rb
```diff
@@ -59,7 +59,7 @@ def run(model, runner, user_arguments)
       return false
     end
 
-    surfaces = processTBD(model, set)
+    io, surfaces = processTBD(model, set)
     surfaces.each do |id, surface|
```
rd2, line 63: For JSON export, the easiest and safest solution is to prune/edit the _io_ hash once the derated _surfaces_ objects have their _:ratio_ values calculated (and therefore right before both objects are returned at the end of processTBD). Initial tests seem to work just fine.

lib/measures/tbd/resources/psi.rb
```diff
@@ -161,7 +161,7 @@ def complete?(s)
   end
 end
 
-def processTBDinputs(surfaces, edges, set, io_path, schema_path)
+def processTBDinputs(surfaces, edges, set, io_path = nil, schema_path = nil)
```
rd2, line 164: As per previous PR request.

spec/tbd_tests_spec.rb
```diff
@@ -1979,6 +1979,12 @@
       next unless id == "Entryway  Wall 5"
       expect(surface[:heatloss]).to be_within(0.01).of(13.96)
     end
+
+    #out = JSON.pretty_generate(io)
+    #out_path = File.dirname(__FILE__) + "/../json/tbd_minimal_KHI.out.json"
+    #File.open(out_path, "w") do |out_path|
+    #  out_path.puts out
+    #end
```
rd2, line 1987: The final version of the _io_ object may be tweaked a few times over the next 48 hours. For now, let's stick its current state.

lib/measures/tbd/resources/psi.rb
macumber: Thanks!

PR #50, opened 2020-12-03
Json tweaks2

macumber:


spec/tbd_tests_spec.rb
```diff
@@ -2428,4 +2428,44 @@
...
     expect(JSON::Validator.validate(schema_path, io_path, uri: true)).to be(true)
   end
+
+  it "has a PSI class" do
+    psi = PSI.new
+    expect(psi.set.has_key?("poor (BC Hydro)")).to be(true)
+    expect(psi.complete?("poor (BC Hydro)")).to be(true)
+
+    expect(psi.set.has_key?("new set")).to be(false)
+    expect(psi.complete?("new set")).to be(false)
```
rd2, line 2438: Good call! Didn't anticipate probing the PSI library with an arbitrary string (bad design).

PR #54, opened 2021-01-02
TBD-generated Kiva settings and objects

rd2:
First draft of TBD generation of Kiva settings and foundation objects.


lib/measures/tbd/measure.rb
```diff
@@ -58,6 +58,12 @@ def arguments(model)
     write_tbd_json.setDefaultValue(true)
     args << write_tbd_json
 
+    gen_kiva = OpenStudio::Measure::OSArgument.makeBoolArgument("gen_kiva", true, false)
+    gen_kiva.setDisplayName("Generate Kiva inputs")
+    gen_kiva.setDescription("Generate OSM Kiva settings and objects if model surfaces have foundation boundary conditions")
+    gen_kiva.setDefaultValue(true)
+    args << gen_kiva
+
     return args
```
rd2, line 67: Checkbox : _checked_ by default, but non-intrusive if no floor surfaces have "Foundation" as outside boundary condition.

lib/measures/tbd/measure.rb
```diff
@@ -92,7 +100,7 @@ def run(model, runner, user_arguments)
       end
     end
 
-    io, surfaces = processTBD(model, option, io_path, schema_path)
+    io, surfaces = processTBD(model, option, io_path, schema_path, gen_kiva)
```
rd2, line 104: New argument to method (true/false).

lib/measures/tbd/measure.rb
```diff
@@ -130,6 +138,76 @@ def run(model, runner, user_arguments)
...
+      # Although one may choose to auto-generate Kiva settings and objects,
+      # there must be at least one valid foundation-facing floor in the model.
+      kiva = false
+      surfaces.values.each do |surface|
+        next if kiva
+        kiva = true if surface.has_key?(:kiva)
+      end
+
+      if kiva
+        arg = "TotalExposedPerimeter"
```
rd2, line 151: "TotalExposedParameter" is the easiest solution for TBD.

lib/measures/tbd/measure.rb
```diff
@@ -130,6 +138,76 @@ def run(model, runner, user_arguments)
...
+        foundation_kiva_settings = model.getFoundationKivaSettings
+        foundation_kiva_settings.setName("TBD-generated Kiva settings template")
```
rd2, line 153: By explicitly declaring the Kiva foundation settings, it provides a placeholder or mechanism for users to (manually) tweak settings.

lib/measures/tbd/measure.rb
```diff
@@ -130,6 +138,76 @@ def run(model, runner, user_arguments)
...
+
+        # Typical circa-1980 slab-on-grade (perimeter) insulation setup.
+        kiva_slab = OpenStudio::Model::FoundationKiva.new(model)
+        kiva_slab.setName("Kiva slab")
+        kiva_slab.setInteriorHorizontalInsulationMaterial(xps_25mm)
+        kiva_slab.setInteriorHorizontalInsulationWidth(0.6)
+
+        # Basement wall setup (full-height insulation as construction layer)
+        kiva_basement = OpenStudio::Model::FoundationKiva.new(model)
+        kiva_basement.setName("Kiva basement")
```
rd2, line 174: If there are foundation walls linked to the foundation slab, then TBD presumes a basement setup. If not, a slab-on-grade setup.

spec/tbd_tests_spec.rb
```diff
@@ -2718,4 +2718,100 @@
...
+    translator = OpenStudio::OSVersion::VersionTranslator.new
+    path = OpenStudio::Path.new(File.dirname(__FILE__) + "/files/midrise_KIVA.osm")
+    os_model = translator.loadModel(path)
+    expect(os_model.empty?).to be(false)
+    os_model = os_model.get
+
+    # Reset all ground-facing surfaces as "foundations".
+    os_model.getSurfaces.each do |s|
+      next unless s.outsideBoundaryCondition.downcase == "ground"
+      s.setOutsideBoundaryCondition("Foundation")
```
macumber, line 2732: I would suggest not setting this here because it puts the model in a temporarily invalid state (because the boundary condition is "Foundation" but there is no foundation object assigned). The boundary condition will be set to "Foundation" when you call `setAdjacentFoundation` and pass in a foundation object below:
https://github.com/NREL/OpenStudio/blob/develop/src/model/Surface.cpp#L1919

Not setting the model to invalid states was always a challenge because often you have to move through invalid states when modifying a model.  It's probably ok to do things as you are because the model is valid at the beginning and end of the measure.  I would recommend against having the user set the boundary condition to "Foundation" in the app before running the measure because that saves the model in the invalid state, which results in EnergyPlus crashing as you found out.

spec/tbd_tests_spec.rb
```diff
@@ -2718,4 +2718,100 @@
...
+      kiva = true if surface.has_key?(:kiva)
+    end
+
+    expect(kiva).to be(true)
+    arg = "TotalExposedPerimeter"
+    foundation_kiva_settings = os_model.getFoundationKivaSettings
+    foundation_kiva_settings.setName("TBD-generated Kiva settings template")
+
+    # Generic 1" XPS insulation.
+    xps_25mm = OpenStudio::Model::StandardOpaqueMaterial.new(os_model)
```
macumber, line 2754: This is a Kiva question, but is this exterior insulation supposed to not be included in the Surface construction too?  Is there a way to check if it is (density < X?) and warn about possible double counting of insulation?

spec/tbd_tests_spec.rb
```diff
@@ -2718,4 +2718,100 @@
...
+        expect(exp).to be_within(0.01).of(19.20) if id == "g GFloor NEA"
+        expect(exp).to be_within(0.01).of(19.20) if id == "g GFloor SWA"
+        expect(exp).to be_within(0.01).of(19.20) if id == "g GFloor SEA"
+        expect(exp).to be_within(0.01).of(11.58) if id == "g GFloor S1A"
+        expect(exp).to be_within(0.01).of(11.58) if id == "g GFloor S2A"
+        expect(exp).to be_within(0.01).of(11.58) if id == "g GFloor N1A"
+        expect(exp).to be_within(0.01).of(11.58) if id == "g GFloor N2A"
+        expect(exp).to be_within(0.01).of(3.36) if id == "g Floor C"
+        s.createSurfacePropertyExposedFoundationPerimeter(arg, surface[:exposed])
+        s.setAdjacentFoundation(kiva_basement) if surface[:kiva] == :basement
```
macumber, line 2798: Might put this in and `if/else` block just to ensure that you will always assign a foundation

spec/tbd_tests_spec.rb
```diff
@@ -2718,4 +2718,100 @@
...
+        s.setAdjacentFoundation(kiva_slab) if surface[:kiva] == :slab
+      end
+
+      # Loop through basement wall surfaces and assign foundation object.
+      surfaces.each do |i, surf|
+        next unless found
+        next unless surf.has_key?(:foundation)
+        next unless id == surf[:foundation]
```
macumber, line 2806: I'm a bit confused about what is going on here.  Basement walls are not tagged with `surface[:kiva] = :basement` but instead the walls reference the basement floor as the `surface[:foundation]`?  This is potentially more confusing because of confusion between tbd foundation object and the kiva foundation object.  Might help to add some comments

spec/tbd_tests_spec.rb
rd2: No, this insulation material shouldn't be included as a construction layer elsewhere (although users can reuse the material in the future - as they see fit). This material is needed for the non-layered insulation configurations illustrated [here](https://github.com/rd2/tbd/blob/kiva/spec/files/KIVA.png).

I should nonetheless do a check first to ensure I'm not overwriting another existing material XPS material.

spec/tbd_tests_spec.rb
rd2: Sure, I could add more comments. TBD surfaces tagged with _:kiva_ are strictly floors with "Foundation" as outside boundary condition (with _:basement_ or _:slab_ as 2x possible values). TBD surfaces with _:foundation_ tags are strictly walls that are part of the same Kiva foundation as the adjacent floor. The _:foundation_ tag in fact holds the name of the foundation floor. It's just handy to quickly identify which walls belong to which foundation/floor.

lib/measures/tbd/README.md
```diff
@@ -43,6 +43,14 @@ Write tbd.out.json to customize for subsequent runs. Edit and place in model fil
 **Required:** true,
 **Model Dependent:** false
 
+### Generate Kiva inputs
+Generate OSM Kiva settings and objects if model surfaces have foundation boundary conditions
```
macumber, line 47: I'd suggest the behavior be that ground surfaces are converted to foundation surfaces rather than asking the user change surfaces to foundation beforehand.

lib/measures/tbd/measure.rb
macumber: I'd suggest the behavior be that ground surfaces are converted to foundation surfaces rather than asking the user change surfaces to foundation beforehand.  You could just do a loop through all ground surfaces and change to foundation at the beginning of your measure.  Is there really a use case to simulate some surfaces with ground boundary condition and others with foundation?

lib/measures/tbd/resources/psi.rb
```diff
@@ -1154,6 +1153,43 @@ def processTBD(os_model, psi_set, io_path = nil, schema_path = nil)
     edge[:set] = p unless psi.empty?
   end                                                                # edge loop
 
+  # Add ":basement" or ":slab" keyword to edge if the following are met:
+  # 1. "foundation"-facing floor surface
+  # 2. 1x foundation-facing (:basement) or exterior-facing (:slab) wall
+  edges.values.each do |edge|
+    next unless gen_kiva
+    next if edge.has_key?(:basement)
```
macumber, line 1161: This is checking if the edge has already been processed?

spec/tbd_tests_spec.rb
rd2: Of course, I'm only setting this up for a small RSpec test - it's not part of the measure per se (so there's no real risk). BTW, not linking a surface with a foundation object does not render the model invalid (EnergyPlus will generate a default foundation object if missing). The issue is with the exposed perimeter (which EnergyPlus will not generate).

I understand the reservations, but the simple fact is that users can currently pick "Foundation" (either in the OpenStudio App or SDK) and therefore generate an invalid model. For the Kiva option to work, users have to select "Foundation" for targeted floors (and optionally walls). Sloped terrain and walkout basements are classic cases where some walls will be tagged as foundation walls wile others will instead be outside-facing. There is no way to make an educated guess, apart from tagging some surfaces with "Foundation". At least the method renders an invalid OSM into a valid one.

lib/measures/tbd/resources/psi.rb
```diff
@@ -1154,6 +1153,43 @@ def processTBD(os_model, psi_set, io_path = nil, schema_path = nil)
...
+    next if edge.has_key?(:slab)
+    edge[:surfaces].keys.each do |id|
+      next if edge.has_key?(:basement)
+      next if edge.has_key?(:slab)
+      next unless floors.has_key?(id)
+      next unless surfaces.has_key?(id)
+      next unless surfaces[id][:boundary].downcase == "foundation"
+
+      # Loop through linked walls: if exterior or foundation ...
+      edge[:surfaces].keys.each do |i|
```
macumber, line 1171: This inner loop is hard for me to follow, I'd suggest some comments for when you have to come back to this in the future

lib/measures/tbd/measure.rb
rd2: "I'd suggest the behavior be that ground surfaces are converted to foundation surfaces ..." 
> ... sounds like a good _--force kiva_ option.

"Is there really a use case to simulate some surfaces with ground boundary condition and others with foundation?"
> ... dunno, maybe.

"... rather than asking the user change surfaces to foundation beforehand."
> ... we're not asking users to do anything. Users currently _want_ to use Kiva but the _exposed perimeter_ input makes it complicated within the OpenStudio context. This Kiva option makes things valid and easier to edit. Not perfect, but better than what's available now.

lib/measures/tbd/resources/psi.rb
rd2: Yes.

lib/measures/tbd/resources/psi.rb
rd2: Noted. There may be leftover code I no longer rely on - will cleanup and comment.

lib/measures/tbd/measure.rb
rd2: The option is in place. Haven't fully tested it yet.

lib/measures/tbd/measure.rb
```diff
@@ -100,6 +108,15 @@ def run(model, runner, user_arguments)
...
     end
 
+    # Process all ground-facing surfaces as foundation-facing.
+    if gen_kiva_force
+      gen_kiva = true
+      model.getSurfaces.each do |s|
+        next unless s.isGroundSurface
+        s.outsideBoundaryCondition == "Foundation"
+      end
+    end
```
rd2, line 118: Straightforward. Haven't fully tested it yet.

lib/measures/tbd/measure.rb
```diff
@@ -138,76 +155,6 @@ def run(model, runner, user_arguments)
...
-              next unless ss.nameString == i
-              next unless ss.outsideBoundaryCondition.downcase == "foundation"
-              ss.setAdjacentFoundation(kiva_basement) if surface[:kiva] == :basement
-              ss.setAdjacentFoundation(kiva_slab) if surface[:kiva] == :slab
-            end
-          end
-        end
-      end
-    end
-
```
rd2: Moved the whole procedure within psi.rb.

lib/measures/tbd/resources/psi.rb
```diff
@@ -542,6 +542,144 @@ def tbdSurfaceEdges(surfaces, edges)
   end
 end
 
+##
```
rd2, line 545: See lib/psi.rb for comments.

lib/measures/tbd/resources/psi.rb
```diff
@@ -1153,42 +1291,9 @@ def processTBD(os_model, psi_set, io_path = nil, schema_path = nil, gen_kiva)
...
-        surfaces[id][:exposed] += edge[:length]
-        surfaces[id][:kiva] = :slab unless surfaces[id].has_key?(:kiva)
-        surfaces[id][:kiva] = :basement if b == "foundation"
-      end
-    end
-  end
-
+  # Generate OSM Kiva settings and objects if foundation-facing floors.
+  # 'kiva' == false if partial failure (log failure eventually).
+  kiva = generateKiva(os_model, walls, floors, edges) if gen_kiva
```
rd2, line 1296: Could report partial failures to runner and/or log messages (to do).

lib/measures/tbd/resources/psi.rb
```diff
@@ -1153,42 +1291,9 @@ def processTBD(os_model, psi_set, io_path = nil, schema_path = nil, gen_kiva)
     edge[:set] = p unless psi.empty?
   end                                                                # edge loop
 
-  # Add ":basement" or ":slab" keyword to edge if the following are met:
```
rd2: Moved procedure to psi.rb.

lib/psi.rb
```diff
@@ -542,6 +542,144 @@ def tbdSurfaceEdges(surfaces, edges)
...
+##
+# Generate OSM Kiva settings and objects if model surfaces have 'foundation'
+# boundary conditions.
+#
+# @param [OpenStudio::Model::Model] os_model An OS model
+# @param [Hash] floors TBD-generated floors
+# @param [Hash] walls TBD-generated walls
+# @param [Hash] edges TBD-generated edges (many linking floors & walls
+#
+# @return [Bool] Returns true if all Kiva foundations are successfully generated.
```
rd2, line 554: New, dedicated procedure.

spec/tbd_tests_spec.rb
```diff
@@ -2652,70 +2584,23 @@
...
     expect(surfaces.size).to eq(180)
 
-    kiva = false
-    surfaces.values.each do |surface|
-      next if kiva
-      kiva = true if surface.has_key?(:kiva)
-    end
+    # Validate.
+    surfaces.each do |id, surface|
+      next unless surface.has_key?(:foundation) # ... only floors
```
rd2, line 2589: Only Kiva floors have _:foundation_ hash keys (corresponding value == actual OS KivaFoundation object).

spec/tbd_tests_spec.rb
```diff
@@ -2794,24 +2647,70 @@
...
+    expect(os_model2.empty?).to be(false)
+    os_model2 = os_model2.get
+
+    os_model2.getSurfaces.each do |s|
+      next unless s.isGroundSurface
+      next unless s.nameString == "Fine Storage" ||
+                  s.nameString == "Office Storage" ||
+                  s.nameString == "Bulk Storage"
+      expect(s.outsideBoundaryCondition).to eq("Foundation")
+    end
```
rd2, line 2710: This works. But if one runs a simulation, EnergyPlus will crash (missing construction name). Could it be that Kiva'ed surfaces can't rely on construction sets?

spec/tbd_tests_spec.rb
rd2: The --force checkbox/argument should accomplish what you suggest. Requires more formal RSpec tests (to do).

spec/tbd_tests_spec.rb
rd2: The most recent commit brought in some sorely needed clarity, so most of the previous comments here no longer apply.

lib/measures/tbd/README.md
rd2: Still needs extensive testing, but the --force Kiva checkbox/argument should do the trick.

lib/measures/tbd/resources/psi.rb
rd2: (this has been revised).

lib/measures/tbd/resources/psi.rb
rd2: The revised code should be clearer.

lib/measures/tbd/resources/psi.rb
```diff
@@ -657,7 +657,7 @@ def generateKiva(os_model, walls, floors, edges)
         # Retrieve surface (standard) construction
         s.setAdjacentFoundation(floors[id][:foundation])
         s.setConstruction(s.construction.get)
-        s.createSurfacePropertyExposedFoundationPerimeter(arg, floors[id][:exposed])
+        s.createSurfacePropertyExposedFoundationPerimeter(arg, floors[id][:exposed]) if floors[id][:exposed] > 0.01
```
rd2, line 660: Kiva doesn't like exposed perimeters == 0. Requires more testing.

lib/psi.rb
```diff
@@ -657,7 +657,9 @@ def generateKiva(os_model, walls, floors, edges)
         # Retrieve surface (standard) construction
         s.setAdjacentFoundation(floors[id][:foundation])
         s.setConstruction(s.construction.get)
-        s.createSurfacePropertyExposedFoundationPerimeter(arg, floors[id][:exposed])
+        if floors[id][:exposed] > 0.01
+          s.createSurfacePropertyExposedFoundationPerimeter(arg, floors[id][:exposed])
```
rd2, line 661: Kiva/EnergyPlus don't like exposed perimeters == 0 (e.g. core zone slab with _foundation_ as outside boundary condition). Requires more testing to circumscribe similar critical cases ...

lib/measures/tbd/measure.rb
rd2: Fixed. Thanks @macumber.

lib/psi.rb
```diff
@@ -654,12 +654,117 @@ def generateKiva(os_model, walls, floors, edges)
...
+        # perimeter object.
+        construction = s.construction.get
         s.setAdjacentFoundation(floors[id][:foundation])
-        s.setConstruction(s.construction.get)
-        if floors[id][:exposed] > 0.01
-          s.createSurfacePropertyExposedFoundationPerimeter(arg, floors[id][:exposed])
-        end
+        s.setConstruction(construction)
+        exp = floors[id][:exposed]
+        exp = 0.01 if exp < 0.01
```
rd2, line 663: This seems to work. The subsequent (lengthy) comments below (embedded in the code) give an idea of what occurs when setting the exposed perimeter to 0.0m or not setting the exposed perimeter at all if 0.0m. I think this points to a bug in OS-to-EPlus translation ...

rd2: See [fix](https://github.com/rd2/tbd/commit/271ba7c25abef6c382f8b42784efde01c727fa5f#r45730611). Thanks for the suggestion @macumber.

lib/measures/tbd/measure.rb
rd2: Both options are now available, and tested.

PR #55, opened 2021-01-06
Fix the measure

macumber:


lib/measures/tbd/measure.rb
```diff
@@ -113,8 +113,9 @@ def run(model, runner, user_arguments)
       gen_kiva = true
       model.getSurfaces.each do |s|
         next unless s.isGroundSurface
-        s.outsideBoundaryCondition == "Foundation"
```
macumber, line 116: @rd2 this was checking equality and not setting the outside boundary condition

rd2: @macumber : thanks! pretty silly really.

PR #60, opened 2021-06-08
Joints

rd2:
Major revision to core functionality and JSON I/O, including:

- rehaul of flat/concave/convex edge variants
- subsequent expansion of inheritance rules
- unconditioned, semi-heated vs conditioned spaces/zones
- handling partially-completed OSMs e.g., no set points, no HVAC
- treatment of party wall edges, (expansion) joints and (mild) transitions
- proportionate derating based on construction RSi-values
- curtain/window wall spandrel PSI sets (suggestions)

I flag below a ~dozen significant changes. Other major enhancements are left uncommented, yet remain well documented in the Ruby code itself.


json/tbd_PSI_combo.json
```diff
@@ -7,8 +7,7 @@
       "rimjoist": 0.7,
       "parapet": 0.7,
       "fenestration": 0.7,
-      "concave": 0.7,
-      "convex": 0.7,
+      "corner": 0.7,
```
rd2, line 10: Published PSI data on _concave_ vs _convex_ variants of a given thermal bridge type (such as corners) remains sparse. Awaiting future published data, the solution here is nonetheless generalized to handle _flat_, _concave_ or _convex_ variants of any thermal bridge type/edge ... with the exception of joints and transitions (discussed a bit later on). Variants may inherit parent PSI values (e.g., _corner_).

json/tbd_PSI_combo.json
```diff
@@ -18,14 +17,13 @@
...
-      "concave": 0.2,
-      "convex": 0.2,
+      "corner": 0.2,
       "balcony": 0.3,
       "party": 0.2,
       "grade": 0.2
     }
   ],
-  "unit": [
+  "building": [
```
rd2, line 26: Less confusing with regards to OpenStudio terminology e.g., a _housing unit_ vs _The building_.

json/tbd_smalloffice.json
```diff
@@ -0,0 +1,28 @@
...
+  "description": "test_smalloffice.osm with building story PSI set",
+  "psis": [
+    {
+      "id": "office PSI set",
+      "rimjoist": 0.300,
+      "parapet": 0.310,
+      "fenestration": 0.320,
+      "cornerconcave": 0.330,
+      "cornerconvex": 0.340,
+      "balcony": 0.350,
```
rd2, line 12: An example of a _concave_ vs _convex_ corner variant.

json/tbd_smalloffice.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "psi": "regular (BETBG)"
+    }
+  ],
+  "stories": [
+    {
+      "id": "Building Story 1",
+      "psi": "office PSI set"
+    }
+  ]
```
rd2, line 27: Expanding the default PSI sets to cover building stories (as does OpenStudio with default construction sets).

json/tbd_warehouse4.json
```diff
@@ -0,0 +1,21 @@
+{
+  "schema": "https://github.com/rd2/tbd/blob/master/tbd.schema.json",
+  "description": "test_warehouse.osm with negative PSI values",
+  "psis": [
+    {
+      "id": "BETBG Detail 7.6.2",
+      "rimjoist": 0.0,
+      "parapet": 0.0,
+      "fenestration": 0.0,
+      "corner": -0.0471,
```
rd2, line 10: Although rare, one will come across negative PSI values in the literature, generally when dealing with efficient thermal breaks while relying on outside measurements (as per A-90.1).

json/tbd_warehouse5.json
```diff
@@ -0,0 +1,41 @@
...
+  "spacetypes": [
+    {
+      "id": "Warehouse Office",
+      "psi": "Warehouse Office"
+    },
+    {
+      "id": "Warehouse Fine",
+      "psi": "Warehouse Fine"
+    }
+  ]
```
rd2, line 40: Default space-type-specific PSI sets, similar to OpenStudio default space-type construction sets.

json/tbd_warehouse6.json
```diff
@@ -0,0 +1,28 @@
...
+        "Fine Storage Office Front Wall"
+      ]
+    },
+    {
+      "psi": "ok",
+      "type": "joint",
+      "surfaces": [
+        "Fine Storage Left Wall",
+        "Fine Storage Office Left Wall"
+      ]
```
rd2, line 25: By default, TBD labels as (mild) _transitions_ edges linking parallel surfaces (e.g. between 2x butting walls sharing same 3D plane, between 2x sloping roof drainage planes). There are 2 of those in the US DOE warehouse prototype building (upper Fine Storage walls, above the office). And also by default, such transitions have a PSI-value of 0 (i.e., no derating occurs). In some cases, such intersections can represent significant thermal bridges, such as structural expansion joints or roof curbs. Users have the option of resetting the thermal bridge type e.g., from _transition_ to _joint_ (or any other type for that matter).

json/tbd_warehouse7.json
```diff
@@ -0,0 +1,35 @@
+{
+  "schema": "https://github.com/rd2/tbd/blob/master/tbd.schema.json",
+  "description": "testing heads, sills & jambs for test_warehouse.osm",
+  "psis": [
+    {
+      "id": "detailed",
+      "head": 0.500,
+      "sill": 0.300,
+      "jamb": 0.100
```
rd2, line 9: Users can either simply rely on a general _fenestration_ thermal bridge type (as before), or differentiate between subsurface _head, sill vs jamb_ edges.

lib/psi.rb
```diff
@@ -13,18 +13,21 @@
 # Set 10mm tolerance for edge (thermal bridge) vertices.
 TOL = 0.01
 
-# Sources for the following defaults KHI & PSI values/sets:
+# Sources for thermal bridge types and/or linked default KHI & PSI values/sets:
 
-# BETB = BC Hydro's Building Envelope Thermal Bridging Guide v1.4
+# BETBG = Building Envelope Thermal Bridging Guide v1.4
 # www.bchydro.com/content/dam/BCHydro/customer-portal/documents/power-smart/
 # business/programs/BETB-Building-Envelope-Thermal-Bridging-Guide-v1-4.pdf
```
rd2, line 20: I'll either revise (if needed) the values based on BETBG v1.5, or wait for v1.6.

lib/psi.rb
```diff
@@ -61,79 +64,256 @@ def append(k)
...
+    # convex corners when using outside dimensions for an OSM.
+    #
+    # PSIe = PSIi + U * 2(Li-Le), where:
+    #   PSIe = adjusted PSI                                          (W/K per m)
+    #   PSIi = initial published PSI                                 (W/K per m)
+    #      U = average clear field U-factor of adjacent walls           (W/m2.K)
+    #     Li = from interior corner to edge of "zone of influence"           (m)
+    #     Le = from exterior corner to edge of "zone of influence"           (m)
+    #
+    #  Li-Le = wall thickness e.g., -0.25m (negative here as Li < Le)
```
rd2, line 103: A placeholder for now. This would need to find its way in future guide material.

lib/psi.rb
```diff
@@ -61,79 +64,256 @@ def append(k)
...
+      parapet:       0.660, # * Detail 1.3.2
+      fenestration:  0.000, # * ... generally part of clear-field RSi
+      corner:        0.200, # * Detail 1.4.2
+      balcony:       0.400, # * Detail 9.1.15
+      party:         0.500, # ** ... similar to parapet/balcony
+      grade:         0.880, # * Detail 2.5.1
+      joint:         0.140, # * Detail 7.4.2
+      transition:    0.000
+    }.freeze               # "good" to "high performance" curtainwall spandrels
+    self.genShorthands("spandrel HP (BETBG)")
```
rd2, line 172: Practical suggestions (there is some need for clarification when dealing with curtain-wall/window-wall technology amongst designers).

lib/psi.rb
```diff
@@ -61,79 +64,256 @@ def append(k)
...
+      h[:gradeconcave]    = @set[p].has_key?(:gradeconcave)
+      h[:gradeconvex]     = @set[p].has_key?(:gradeconvex)
+      h[:balcony]         = @set[p].has_key?(:balcony)
+      h[:balconyconcave]  = @set[p].has_key?(:balconyconcave)
+      h[:balconyconvex]   = @set[p].has_key?(:balconyconvex)
+      h[:rimjoist]        = @set[p].has_key?(:rimjoist)
+      h[:rimjoistconcave] = @set[p].has_key?(:rimjoistconcave)
+      h[:rimjoistconvex]  = @set[p].has_key?(:rimjoistconvex)
+      @has[p] = h
+
```
rd2, line 242: "@has" : True/False response as to whether a PSI set contains a given thermal bridge type. Relying on Topolys, TBD tags an edge with one (or more) of the above labels. This is systematically populated once a PSI set is added and held in memory for quick access.

lib/psi.rb
```diff
@@ -61,79 +64,256 @@ def append(k)
...
+      v[:balconyconvex]   = @set[p][:balconyconvex]   if h[:balconyconvex]
+      v[:rimjoist]        = @set[p][:rimjoist]        if h[:rimjoist]
+      v[:rimjoistconcave] = @set[p][:rimjoist]        if h[:rimjoist]
+      v[:rimjoistconvex]  = @set[p][:rimjoist]        if h[:rimjoist]
+      v[:rimjoistconcave] = @set[p][:rimjoistconcave] if h[:rimjoistconcave]
+      v[:rimjoistconvex]  = @set[p][:rimjoistconvex]  if h[:rimjoistconvex]
+
+      max = [v[:parapetconcave], v[:parapetconvex]].max
+      v[:parapet] = max unless @has[:parapet]
+      @val[p] = v
```
rd2, line 315: Class member _val_ holds PSI values (in W/K per meter) for all possible variants. As with _has_, this is populated once and held in memory for quick access.

lib/psi.rb
```diff
@@ -416,28 +722,612 @@ def processTBDinputs(surfaces, edges, set, io_path = nil, schema_path = nil)
...
   return io, psi, khi
 end
 
+##
+# Return min & max values for schedule (ruleset).
+#
+# @param [OpenStudio::Model::ScheduleRuleset] sched An OS schedule (ruleset)
+#
+# @return [Hash] :min & :max; nilled if invalid.
+def scheduleRulesetMinMax(sched)
```
rd2, line 754: This, and several other related functions, allow TBD to distinguish unenclosed/unconditioned vs semi-heated/conditioned spaces/thermal zones, attics, plenums, etc., IF required inputs are present in an OSM.

lib/psi.rb
```diff
@@ -469,72 +1358,136 @@ def transforms(model, group)
...
+
+##
+# Validate whether edge surfaces form a concave angle, as seen from outside.
+#
+# @param [Surface] s1 A first TBD surface
+# @param [Surface] s2 A second TBD surface
+#
+# @return [Bool] Returns true if angle between surfaces is concave
+def concave?(s1, s2)
+  concave = false
```
rd2, line 1382: General solution, no longer reserved for corners.

json/tbd_warehouse6.json
macumber: I'm assuming that any explicitly listed edge types will override the automatically deduced type?  Basically, you first compute the deduced type and then go through and override with any types the user sets right?

macumber: This should be one of the features of the UI tool.  Click on an edge, see what type is assigned.  If you don't like it, maybe the UI tool could offer to create an override for you?

json/tbd_warehouse7.json
macumber: Do window edges derate the parent surface or the window?  Maybe it is an option?  Can the user decide to do advanced modeling of the window using https://windows.lbl.gov/software/window and tell TBD not to derate head, sill, and jambs?  Would they just put in 1.0 for the psi factors?

json/midrise.json
```diff
@@ -0,0 +1,68 @@
+{
+  "schema": "https://github.com/rd2/tbd/blob/master/tbd.schema.json",
+  "description": "sorting out multiple story PSI sets (midrise)",
+  "psis": [
+    {
+      "id": "compliant",
+      "rimjoist": 0.3,
+      "parapet": 0.325,
+      "fenestration": 0.35,
+      "corner": 0.45,
```
macumber, line 10: If a model has a concave corner but the PSI set only has corner, will that edge fall back to use corner?

lib/measures/tbd/resources/psi.rb
```diff
@@ -61,79 +64,256 @@ def append(k)
...
+      transition:    0.000
+    }.freeze               # "conventional", closer to window wall spandrels
+    self.genShorthands("spandrel (BETBG)")
+
+    @set["spandrel HP (BETBG)"] =
+    {
+      rimjoist:      0.170, # * Detail 1.2.7
+      parapet:       0.660, # * Detail 1.3.2
+      fenestration:  0.000, # * ... generally part of clear-field RSi
+      corner:        0.200, # * Detail 1.4.2
```
macumber, line 165: No concavecorner?

lib/measures/tbd/resources/psi.rb
```diff
@@ -61,79 +64,256 @@ def append(k)
...
+      v[:transition]      = @set[p][:transition]      if h[:transition]
+      v[:fenestration]    = @set[p][:fenestration]    if h[:fenestration]
+      v[:head]            = @set[p][:fenestration]    if h[:fenestration]
+      v[:headconcave]     = @set[p][:fenestration]    if h[:fenestration]
+      v[:headconvex]      = @set[p][:fenestration]    if h[:fenestration]
+      v[:sill]            = @set[p][:fenestration]    if h[:fenestration]
+      v[:sillconcave]     = @set[p][:fenestration]    if h[:fenestration]
+      v[:sillconvex]      = @set[p][:fenestration]    if h[:fenestration]
+      v[:jamb]            = @set[p][:fenestration]    if h[:fenestration]
+      v[:jambconcave]     = @set[p][:fenestration]    if h[:fenestration]
```
macumber, line 265: Ok, I think I see how this works.  First jambconcave gets fenestration, then jamb (if it exists), then jambconcave (if that exists)

lib/measures/tbd/resources/psi.rb
```diff
@@ -142,66 +322,162 @@ def initialize
...
+    return false unless @val.has_key?(p)
+    raise "missing @has key #{p}" unless @has.has_key?(p)
+    raise "missing head" unless @has[p].has_key?(:head)
+    holes = []
+    holes << :head if @has[p][:head]
+    holes << :sill if @has[p][:sill]
+    holes << :jamb if @has[p][:jamb]
+    ok = holes.size == 3
+    ok = true if @has[p][:fenestration]
+    return false unless ok
```
macumber, line 409:
```suggestion
    ok = holes.size == 3
    return false unless ok
    ok = true if @has[p][:fenestration]
    return false unless ok
```

macumber: I think the first assignment to ok gets overridden?

lib/measures/tbd/resources/psi.rb
```diff
@@ -416,28 +722,612 @@ def processTBDinputs(surfaces, edges, set, io_path = nil, schema_path = nil)
...
     raise "Incomplete PSI set #{set}" unless psi.complete?(set)
-    io[:unit] = [{ psi: set }]                 # i.e. default PSI set & no KHI's
+    io[:building] = [{ psi: set }]             # i.e. default PSI set & no KHI's
   end
-
   return io, psi, khi
 end
 
+##
+# Return min & max values for schedule (ruleset).
```
macumber, line 749: Wow you really had to add a lot to get the conditioning status.  You might consider putting this in a separate file?

macumber:
This is a huge pull request, I can tell there is a ton of work in here, nice job!  I can't fully review all of the algorithm work in this, I could look at anything in particular if you have something you are less sure about.  I did see that you added tests which is great.  You might consider looking at https://github.com/relevance/rcov to compute code coverage for your tests?  It might help you find some code paths that are not covered.

json/tbd_warehouse6.json
rd2: You got it. This is very similar to how OpenStudio manages surface constructions: use the building default set (if defined); use instead the story set if applicable; use instead the spacetype set if applicable, etc. down to the individual surface construction (overriding all previous default sets). TBD goes down to the individual edge.

json/tbd_warehouse7.json
rd2: Great bunch of questions grouped together (a good reminder of what's needed in user guide material). Firstly, TBD never derates subsurfaces, as per BETBG guidelines (which I quite agree with). Doing so would muddy the waters of what falls into the scope of relevant fenestration standards such as NFRC 100 and CSA A440.2, and how they're referenced in codes. These standards already factor-in the effect of frames/mullions/dividers in the calculation of the overall transmittance/conductance of the fenestrated product per se. The [frame/divider](https://openstudiocoalition.org/tutorials/tutorial_windowproperty_frameanddivider/) option (which is great) is an extension of this. What the standards (and the frame/divider option) do not offer is PSI data on how to derate host surfaces (where thermal bridging is mainly linked to the concentration of supporting structural elements around the opening e.g., lintels). Although TBD relies on similar terminology (e.g., jambs vs heads), it is only interested in the brute subsurface opening edges (shared with the parent surface). So there's no overlap.

rd2: Side note: if one wants to cancel the thermal bridging effect of an edge in TBD, one specifies "0.000" (W/K per m) as a PSI value - not "1.000" (the latter suggests a huge thermal bridge e.g., a cantilevered balcony).

rd2: This [statement](https://bigladdersoftware.com/epx/docs/8-2/input-output-reference/page-011.html#windowpropertyframeanddivider) "The window vertices that you specify in the FenestrationSurface:Detailed object are those of the glazed part of the window, not the frame. EnergyPlus automatically subtracts the area of the frame–determined from the glazing dimensions and the frame width–from the area of the wall containing the window." ... makes me pause. I need to dive into this rabbit hole, I'm afraid. If in some [cases](https://bigladdersoftware.com/epx/docs/8-4/input-output-reference/group-surface-construction-elements.html#windowmaterialsimpleglazingsystem) the OSM vertices are defining "outer" edges of a subsurface, while they may be defining the "inner" edges of a subsurface elsewhere, then TBD needs suitable logic/rules to identify which cases it's processing, and adapt. I'm assuming that there must be some API function that reports overall subsurface area (regardless of whether inner or outer edges are defined) - this would be required to consistently determine WWR.

json/midrise.json
rd2: Yup.

lib/measures/tbd/resources/psi.rb
rd2: In this case, TBD will tag corner edges in an OSM as either _concave_ or _convex_ - both of which will inherit the simple _corner_ PSI value of 0.200 W/K per m. If one has access to PSI data for a given construction system, differentiating _concave_ vs _convex_ corners (e.g. 0.150 vs 0.250), then one can add _cornerconcave_ and _cornerconvex_ items in the PSI set.

lib/measures/tbd/resources/psi.rb
rd2: You got it.

lib/measures/tbd/resources/psi.rb
rd2: Yes, but in both cases it's returning "true" (which is what I'm after). I think in your suggestion, it would return "false" (line 408) if it doesn't have 3 items. The bulk of the tested PSI sets only have fenestration items, yet TBD will systematically generate head, sill and jamb edges. I'll expand the test suite to illustrate this.

lib/measures/tbd/resources/psi.rb
rd2: That is a good idea. I can see practical uses for this outside of TBD. Thanks.

rd2:
Thanks @macumber. I take note of the rcov suggestion. If I had 10 hours, I'd first expand the test suite to cover more unconventional building geometries (i.e. not the US DOE prototypes). The Scrigno example is a good test. I have an even more extreme OSM on hand (à la Gehry or Hadid designs). Otherwise, there's the fenestration frame/glazing rabbit hole ...

spec/tbd_tests_spec.rb
```diff
@@ -2637,6 +4196,105 @@
...
+    expect(psi.complete?("no sills")).to be(true)        # can be a building set
+    holds, vals = psi.shorthands("no sills")
+    expect(holds.empty?).to be(false)
+    expect(vals.empty?).to be(false)
+    expect(holds[:sill]).to be(false)                      # absent from PSI set
+    expect(holds[:sillconcave]).to be(false)               # absent from PSI set
+    expect(holds[:sillconvex]).to be(false)                # absent from PSI set
+    expect(vals[:sill]).to        be_within(0.001).of(0.391)
+    expect(vals[:sillconcave]).to be_within(0.001).of(0.391)
+    expect(vals[:sillconvex]).to  be_within(0.001).of(0.391)     # :fenestration
```
rd2, line 4297: The _complete?_ method always returns "true", whether PSI sets are holding all, some or no _sill_ items. The _holds_ hash correctly reports what the PSI set carries, while _vals_ returns inherited PSI values.

lib/measures/tbd/resources/psi.rb
rd2: Done. See conditioned.rb. Piece of cake. Thanks for the suggestion.

spec/tbd_tests_spec.rb
```diff
@@ -2637,6 +4196,251 @@
...
+    #   else
+    #     heatloss_from_edge = edge[:length] x PSI
+    #   end
+    #
+    # A similar workaround will be needed for API-reported building WWR.
+    #
+    # ... subsurface.netArea calculation here could be reconsidered (?):
+    # https://github.com/NREL/OpenStudio/blob/
+    # 70a5549c439eda69d6c514a7275254f71f7e3d2b/src/model/Surface.cpp#L1446
+    os_model_FD.save("os_model_FD.osm", true)
```
rd2, line 4443: @macumber: I'm hoping these in-line comments are sufficient to explain the API-reported netArea issues when relying on the Frame & Divider objects. If you have a few minutes, could you go over my reasoning/tests?. The workarounds for TBD (corrected edge lengths, netArea) are easy enough - but the netArea reporting could be revised in the API (IMO). Thanks.

json/tbd_warehouse7.json
rd2: @macumber : see [here](https://github.com/rd2/tbd/pull/60#discussion_r649312578).

lib/framedivider.rb
```diff
@@ -0,0 +1,305 @@
...
+  # Should convert (annoying) 4-point subsurface into triangle ...
+  #        A
+  #       / \
+  #      /   \
+  #     /     \
+  #    B - C - D   <<< allowed as OpenStudio/E+ subsurface?
+  #
+  width = 0
+  if s.windowPropertyFrameAndDivider.empty?
+    return area, points
```
rd2, line 93: No Frame & Divider object? Return original subsurface gross area (or net, same thing) and _transformed_ (otherwise unaltered) points.

lib/framedivider.rb
```diff
@@ -0,0 +1,305 @@
...
+  pts[:A] = {}
+  pts[:B] = {}
+  pts[:C] = {}
+  pts[:D] = {} if four
+
+  pts[:A][:pt] = points[0]
+  pts[:B][:pt] = points[1]
+  pts[:C][:pt] = points[2]
+  pts[:D][:pt] = points[3] if four
+
```
rd2, line 111: The step-by-step procedure here is quite pedantic - goal is to clearly articulate what is required. This could be collapsed into 30 lines of code.

lib/framedivider.rb
```diff
@@ -0,0 +1,305 @@
...
+  # Validate if rough opening 'fits' within host surface (think acute angles).
+  # ref: https://github.com/NREL/OpenStudio/blob/
+  #              e603410ac3567d44e7660294f3b06a8a2d591118/ruby/test/
+  #              Polygon3d_PointInPolygonUp_Test.rb
+  vec = OpenStudio::Point3dVector.new
+  dad.each do |vertex|
+    vec.push(OpenStudio::Point3d.new(vertex.x, vertex.y, vertex.z))
+  end
+  # polygon = OpenStudio::Polygon3d.new(vec) <<< error : uninitialized constant OpenStudio::Polygon3d
```
rd2, line 279: @macumber  : unsuccessful initializing Polygon3D (what am I not _requiring_ as a Ruby gem?)

lib/measures/tbd/resources/psi.rb
```diff
@@ -962,25 +1392,64 @@ def processTBD(os_model, psi_set, io_path = nil, schema_path = nil, gen_kiva)
...
-      net:      s.netArea,
-      points:   points,
-      minz:     minz,
-      n:        n
+      type:         type,
+      conditioned:  conditioned,
+      ground:       ground,
+      boundary:     boundary,
+      space:        space,
+      gross:        s.grossArea,
```
rd2, line 1405: Only populating TBD opaque surfaces with API-reported gross area. For net area, relying on Frame & Divider-related workarounds.

lib/measures/tbd/resources/psi.rb
```diff
@@ -993,17 +1462,17 @@ def processTBD(os_model, psi_set, io_path = nil, schema_path = nil, gen_kiva)
     t, r = transforms(os_model, space)
     n = trueNormal(s, r)
 
+    gross, points = opening(os_model, id, t)
```
rd2, line 1465: Subsurface rough opening area & pre-Topolys points from Frame & Divider workarounds.

lib/measures/tbd/resources/psi.rb
```diff
@@ -1085,12 +1554,24 @@ def processTBD(os_model, psi_set, io_path = nil, schema_path = nil, gen_kiva)
...
+    end
+    if p.has_key?(:doors)
+      p[:doors].values.each     do |o| subarea += o[:gross]; end
+    end
+    if p.has_key?(:skylights)
+      p[:skylights].values.each do |o| subarea += o[:gross]; end
+    end
+    p[:net] = p[:gross] - subarea
+  end
+
```
rd2, line 1571: An opaque surface's net area is no longer API-reported - relying on Frame & Divider workarounds.

spec/tbd_tests_spec.rb
rd2: @macumber : See framedivider.rb for workarounds.

spec/tbd_tests_spec.rb
```diff
@@ -2637,6 +4210,530 @@
...
+    surfaces[nom][:windows].each do |i, window|
+      expect(window.has_key?(:points)).to be(true)
+      expect(window[:points].size).to eq(4)
+      if i == name
+        expect(window.has_key?(:gross)).to be(true)
+        expect(window[:gross]).to be_within(0.01).of(5.89)          # ! API 5.58
+      end
+    end
+
+
```
rd2, line 4492: A visual of the detailed examples below : spec/files/TBD_FD.png

json/tbd_warehouse7.json
rd2: The Frame & Divider (framedivider.rb) workarounds address these issues (it was worth diving into the rabbit hole). Ideally, this should eventually make its way into the OpenStudio API.

tbd.gemspec
```diff
@@ -37,9 +37,20 @@ Gem::Specification.new do |spec|
...
+    spec.add_development_dependency "bundler",        "~> 2.1"
+    spec.add_development_dependency "public_suffix",  "~> 3.1.1"
+    spec.add_development_dependency "json-schema",    "~> 2.7.0"
+    spec.add_development_dependency "parallel",       "~> 1.19.2"
+    spec.add_development_dependency "rake",           "~> 13.0"
+    spec.add_development_dependency "rspec",          "~> 3.9"
+    spec.add_development_dependency "rubocop",        "~> 0.54.0"
+    spec.add_development_dependency "yard",           "~> 0.9"
+  else
+    spec.required_ruby_version = "~> 2.7.0"
```
macumber, line 52: This handles OpenStudio 3.2

lib/psi.rb
```diff
@@ -1,11 +1,11 @@
 require "openstudio"
-require "conditioned"
-require "framedivider"
 
 begin
   # try to load from the gem
   require "topolys"
 rescue LoadError
+  require_relative "conditioned"
+  require_relative "framedivider"
```
rd2, line 8: (silly me).

Rakefile
```diff
@@ -10,7 +10,7 @@ task default: :spec
 
 require 'yard'
 YARD::Rake::YardocTask.new do |t|
-  t.files   = ["lib/psi.rb"]
+  t.files   = ["lib/psi.rb", "lib/conditioned.rb", "lib/framedivider.rb"]
```
rd2, line 13: I haven't tested this yet ...

lib/framedivider.rb
```diff
@@ -1,5 +1,81 @@
 require "openstudio"
 
+
+##
+# Flatten OpenStudio 3D points vs Z-axis (Z=0).
+#
+# @param [Array] pts Point3D array
+#
+# @return [Array] Flattened OS 3D points array
+def flatZ(pts)
```
rd2, line 10: Forces Z-coordinates (=0), suppressing non-coplanar flagging (_pointInPolygon_).

lib/framedivider.rb
```diff
@@ -1,5 +1,81 @@
...
+
+##
+# Validate whether an OpenStudio subsurface can 'fit in' (vs parent, siblings)
+#
+# @param [OpenStudio::Model::Model] model An OS model
+# @param [String] id SubSurface identifier
+# @param [Array] pts Point3D array of subsurface 'id'
+#
+# @return [Bool] Returns true if subsurface can fit in.
+def fit?(model, id, pts)
```
rd2, line 33: Entirely an OpenStudio-specific solution now i.e., no Topolys. Doesn't need space transformations either.

lib/framedivider.rb
```diff
@@ -1,5 +1,81 @@
...
+    siblings << flatZ( (ft * sub.vertices).reverse )
+  end
+
+  ft_pts = flatZ(ft * pts)
+  ft_pts.each do |ft_pt|
+    return false unless OpenStudio::pointInPolygon(ft_pt, ft_parent, TOL)
+    siblings.each do |sibling|
+      return false if OpenStudio::pointInPolygon(ft_pt, sibling, TOL)
+    end
+  end
```
rd2, line 76: Also checking if new _rough opening_ vertices are overlapping any of the other subsurfaces (i.e., siblings).

lib/framedivider.rb
```diff
@@ -107,10 +167,10 @@ def opening(model, id, t)
...
   pts[:D] = {} if four
 
-  pts[:A][:pt] = points[0]
-  pts[:B][:pt] = points[1]
-  pts[:C][:pt] = points[2]
-  pts[:D][:pt] = points[3] if four
+  pts[:A][:pt] = Topolys::Point3D.new(points[0].x, points[0].y, points[0].z)
+  pts[:B][:pt] = Topolys::Point3D.new(points[1].x, points[1].y, points[1].z)
+  pts[:C][:pt] = Topolys::Point3D.new(points[2].x, points[2].y, points[2].z)
+  pts[:D][:pt] = Topolys::Point3D.new(points[3].x, points[3].y, points[3].z) if four
```
rd2, line 173: I state earlier that these functions are entirely OpenStudio - I lied. The new _rough opening_ vertices are estimated here from Topolys vector manipulations, then reconverted back to OpenStudio 3D points. In due course, I can recode to ensure a 100% OpenStudio solution.

lib/measures/tbd/resources/psi.rb
```diff
@@ -1458,11 +1459,13 @@ def processTBD(os_model, psi_set, ioP = nil, schemaP = nil, g_kiva = false)
...
     id    = s.nameString
 
+    gross, pts = opening(os_model, id)
+
     # Site-specific (or absolute, or true) surface normal.
     t, r = transforms(os_model, space)
     n = trueNormal(s, r)
 
-    gross, points = opening(os_model, id, t)
+    points = (t * pts).map{ |v| Topolys::Point3D.new(v.x, v.y, v.z) }
```
rd2, line 1468: Topolys subsurface points are only populated (post-transformation) once new _rough opening_ vertices have been pre-validated by the _opening()_ function (entirely an OpenStudio-based operation).

lib/psi.rb
```diff
@@ -1,17 +1,18 @@
...
-  require_relative "conditioned"
-  require_relative "framedivider"
   require_relative "geometry.rb"
   require_relative "model.rb"
   require_relative "transformation.rb"
   require_relative "version.rb"
 end
 
+require_relative "conditioned.rb"
+require_relative "framedivider.rb"
```
rd2, line 14: Simple fix (_conditioned_ & _framedivider_ are not Topolys resources).

PR #63, opened 2021-07-04
Logger

rd2:
Ready for a PR review.

### DIFF
--- a/json/midrise.json
+++ b/json/midrise.json
@@ -1,5 +1,5 @@
 {
-  "schema": "https://github.com/rd2/tbd/blob/master/tbd.schema.json",
+  "schema": "https://github.com/rd2/tbd/blob/logger/tbd.schema.json",
   "description": "sorting out multiple story PSI sets (midrise)",
   "psis": [
     {
@@ -60,9 +60,7 @@
       "psi": "Building Story 3"
     }
   ],
-  "building": [
-    {
-      "psi": "compliant"
-    }
-  ]
+  "building": {
+    "psi": "compliant"
+  }
 }
--- a/json/midrise.json
+++ b/json/midrise.json
@@ -1,8 +1,7 @@
 {
   "schema": "https://github.com/rd2/tbd/blob/logger/tbd.schema.json",
   "description": "sorting out multiple story PSI sets (midrise)",
-  "psis": [
-    {
+  "psis": [{
       "id": "compliant",
       "rimjoist": 0.3,
       "parapet": 0.325,
@@ -46,8 +45,10 @@
       "grade": 0.371
     }
   ],
-  "stories": [
-    {
+  "building": {
+    "psi": "compliant"
+  },
+  "stories": [{
       "id": "Building Story 1",
       "psi": "Building Story 1"
     },
@@ -59,8 +60,5 @@
       "id": "Building Story 3",
       "psi": "Building Story 3"
     }
-  ],
-  "building": {
-    "psi": "compliant"
-  }
+  ]
 }
--- a/lib/conditioned.rb
+++ b/lib/conditioned.rb
@@ -102,7 +102,8 @@ def scheduleRulesetMinMax(sched)
   # standards/Standards.ScheduleRuleset.rb#L124
   result = { min: nil, max: nil }
   unless sched && sched.is_a?(OpenStudio::Model::ScheduleRuleset)
-    TBD.log(TBD::DEBUG, "Invalid sched - ruleset MinMax")
+    TBD.log(TBD::DEBUG,
+      "Invalid ruleset MinMax schedule (argument) - skipping")
     return result
   end
 
@@ -149,7 +150,8 @@ def scheduleConstantMinMax(sched)
   # standards/Standards.ScheduleConstant.rb#L21
   result = { min: nil, max: nil }
   unless sched && sched.is_a?(OpenStudio::Model::ScheduleConstant)
-    TBD.log(TBD::DEBUG, "Invalid sched - constant MinMax")
+    TBD.log(TBD::DEBUG,
+      "Invalid constant MinMax schedule (argument) - skipping")
     return result
   end
 
@@ -177,7 +179,8 @@ def scheduleCompactMinMax(sched)
   # standards/Standards.ScheduleCompact.rb#L8
   result = { min: nil, max: nil }
   unless sched && sched.is_a?(OpenStudio::Model::ScheduleCompact)
-    TBD.log(TBD::DEBUG, "Invalid sched - compact MinMax")
+    TBD.log(TBD::DEBUG,
+      "Invalid compact MinMax schedule (argument) - skipping")
     return result
   end
 
@@ -223,7 +226,8 @@ def maxHeatScheduledSetpoint(zone)
   setpoint = nil
   dual = false
   unless zone && zone.is_a?(OpenStudio::Model::ThermalZone)
-    TBD.log(TBD::DEBUG, "Invalid zone - max T")
+    TBD.log(TBD::DEBUG,
+      "Invalid max heat setpoint thermal zone (argument) - skipping")
     return setpoint, dual
   end
 
@@ -394,7 +398,8 @@ end
 def heatingTemperatureSetpoints?(model)
   answer = false
   unless model && model.is_a?(OpenStudio::Model::Model)
-    TBD.log(TBD::DEBUG, "Invalid model - heat setpoints?")
+    TBD.log(TBD::DEBUG,
+      "Can't find or validate OSM (argument) for heating setpoints - skipping")
     return answer
   end
 
@@ -422,7 +427,8 @@ def minCoolScheduledSetpoint(zone)
   setpoint = nil
   dual = false
   unless zone && zone.is_a?(OpenStudio::Model::ThermalZone)
-    TBD.log(TBD::DEBUG, "Invalid zone - minT")
+    TBD.log(TBD::DEBUG,
+      "Invalid min cool setpoint thermal zone (argument) - skipping")
     return setpoint, dual
   end
 
@@ -579,7 +585,8 @@ end
 def coolingTemperatureSetpoints?(model)
   answer = false
   unless model && model.is_a?(OpenStudio::Model::Model)
-    TBD.log(TBD::DEBUG, "Invalid model - cool setpoints?")
+    TBD.log(TBD::DEBUG,
+      "Can't find or validate OSM (argument) for cooling setpoints - skipping")
     return answer
   end
 
@@ -600,7 +607,8 @@ end
 def airLoopsHVAC?(model)
   answer = false
   unless model && model.is_a?(OpenStudio::Model::Model)
-    TBD.log(TBD::DEBUG, "Invalid model - loops?")
+    TBD.log(TBD::DEBUG,
+      "Can't find or validate OSM (argument) for HVAC air loops - skipping")
     return answer
   end
 
@@ -636,10 +644,19 @@ def plenum?(space, loops, setpoints)
   #           "inactive" thermostat (i.e., can't extract valid setpoints); or
   #   case C. spacetype is "plenum".
   cl = OpenStudio::Model::Space
-  unless space && loops && setpoints && space.is_a?(cl) &&
-         (loops == true || loops == false ) &&
-         (setpoints == true || setpoints == false)
-    TBD.log(TBD::DEBUG, "Invalid arguments - plenum?")
+  unless space && space.is_a?(cl)
+    TBD.log(TBD::DEBUG,
+      "Invalid plenum space (argument) - skipping")
+    return false
+  end
+  unless loops == true || loops == false
+    TBD.log(TBD::DEBUG,
+      "Invalid plenum loops (argument) - skipping")
+    return false
+  end
+  unless setpoints == true || setpoints == false
+    TBD.log(TBD::DEBUG,
+      "Invalid plenum setpoints (argument) - skipping")
     return false
   end
 
--- a/lib/conditioned.rb
+++ b/lib/conditioned.rb
@@ -101,16 +101,15 @@ def scheduleRulesetMinMax(sched)
   # 99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/
   # standards/Standards.ScheduleRuleset.rb#L124
   result = { min: nil, max: nil }
-  raise "Invalid sched (ruleset MinMax)" unless sched
-  cl = OpenStudio::Model::ScheduleRuleset
-  raise "#{sched.class}? expected #{cl} (ruleset)" unless sched.is_a?(cl)
+  unless sched && sched.is_a?(OpenStudio::Model::ScheduleRuleset)
+    TBD.log(TBD::DEBUG, "Invalid sched - ruleset MinMax")
+    return result
+  end
 
   profiles = []
   profiles << sched.defaultDaySchedule
   rules = sched.scheduleRules
-  rules.each do |rule|
-    profiles << rule.daySchedule
-  end
+  rules.each { |rule| profiles << rule.daySchedule }
 
   min = nil
   max = nil
@@ -149,9 +148,10 @@ def scheduleConstantMinMax(sched)
   # 99cf713750661fe7d2082739f251269c2dfd9140/lib/openstudio-standards/
   # standards/Standards.ScheduleConstant.rb#L21
   result = { min: nil, max: nil }
-  raise "Invalid sched (constant MinMax)" unless sched
-  cl = OpenStudio::Model::ScheduleConstant
-  raise "#{sched.class}? expected #{cl} (constant)" unless sched.is_a?(cl)
+  unless sched && sched.is_a?(OpenStudio::Model::ScheduleConstant)
+    TBD.log(TBD::DEBUG, "Invalid sched - constant MinMax")
+    return result
+  end
 
   min = nil
   min = sched.value if sched.value.is_a?(Numeric)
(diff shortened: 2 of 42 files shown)


lib/measures/tbd/measure.rb
```diff
@@ -15,6 +15,36 @@
...
+    File.open(out_path, "w") do |l|
+      l.write msgs.join("\n")
+      begin
+        l.fsync
+      rescue StandardError
+        l.flush
+      end
+    end
+
+    io[:logs] = msgs if io
```
rd2, line 46: @macumber : The log messages (first written out to "tbd.log") are also repeated in the "tbd.out.json". This seems right for 3rd-party apps reading-in TBD JSON output, or for code compliance checks. More practical than dragging 2x files instead of 1x. This now raises the question why even bother generating a "tbd.log" file in the first place, if everything can be stored in a single "tbd.out.json" file? It may even be worth time-stamping/logging "success" messages as well (e.g., the last time TBD info was "successfully" generated). Thoughts?

spec/tbd_tests_spec.rb
```diff
@@ -4869,7 +5155,267 @@
...
+    #     {
+    #       "psi": "detailed",                          <<<< non-fatal ERROR # 3
+    #       "type": "fenestration",
+    #       "surfaces": [
+    #         "Office Front Wall",
+    #         "Office Front Wall Window 1"
+    #       ]
+    #     }
+    #   ]
+    # }
```
rd2, line 5307: In addition to the preceding OSM anomalies introduced earlier (e.g., borderline RSi values, 5-sided windows), pointing out here the invalid TBD JSON entries that will trigger warnings or errors.

spec/tbd_tests_spec.rb
```diff
@@ -4869,7 +5155,267 @@
...
+
+    expect(TBD.status).to eq(TBD::ERROR)
+    expect(TBD.logs.size).to eq(6)
+    # TBD.logs.each { |log| puts log[:msg] }
+    #   'clerestory' vertex count (3 or 4)
+    #   Can't override 'regular (BETBG)' PSI set  - skipping
+    #   'Office Front Wall' KHI 'beam' mismatch
+    #   'Office Front Wall' edge PSI set mismatch - skipping
+    #   Can't assign 180.007 W/K to 'Office Left Wall' - too conductive
+    #   Can't derate 'Fine Storage Right Wall' - material already derated
```
rd2, line 5332: In normal circumstances, and as noted earlier, these log messages will make their way to both "tbd.log" & "tbd.out.json" files (enriched with their recorded timestamps and respective log levels e.g., "ERROR"). Thinking of changing the JSON schema to conserve _key:value_ pairs for :time, :log_level &. :messages (might as well make it easier for 3rd-party applications to parse logged information). @macumber : thoughts?

lib/measures/tbd/measure.rb
macumber: I like the idea of generating a single `tbd.out.json`, that would be easiest for a third party app like SIMEB to parse.  They could check for errors when they open the results.

rd2: Agree - working on it. Why not take one step further: why not systematically generate a `tbd.out.json` under _/files_ in all circumstances? At a minimum, it would contains a higher level, timestamped success/failure message e.g., "Success, yet raised warnings", warning/error logs if applicable, plus the usual runner [output](https://github.com/rd2/tbd/blob/1ddd663ecb5938c2ff1d81f9c919955b4dce1151/lib/measures/tbd/measure.rb#L131) if successful. I'll update the PR for you to test.

lib/measures/tbd/measure.rb
```diff
@@ -18,32 +18,118 @@
...
+
+  tbd_log = { date: Time.now, status: status }
+
+  tbd_msgs = []
+  TBD.logs.each do |l|
+    tbd_msgs << { level: TBD.tag(l[:level]), message: l[:message] }
+    if l[:level] > TBD::INFO
+      runner.registerWarning("(#{TBD.tag(l[:level])}) #{l[:message]}")
+    else
+      runner.registerInfo("(#{TBD.tag(l[:level])}) #{l[:message]}")
```
rd2, line 73: Early in _exitTBD()_, FATAL errors are registered as _warnings_. TBD & OpenStudio are nonetheless halted at the very end of _exitTBD_ (if FATAL).

lib/measures/tbd/measure.rb
```diff
@@ -18,32 +18,118 @@
...
+
+  # User's may not be requesting detailed output - delete non-essential items.
+  io.delete(:psis)        unless out
+  io.delete(:khis)        unless out
+  io.delete(:building)    unless out
+  io.delete(:stories)     unless out
+  io.delete(:spacetypes)  unless out
+  io.delete(:spaces)      unless out
+  io.delete(:surfaces)    unless out
+  io.delete(:edges)       unless out
```
rd2, line 87: Pruning _io_ if user prefers a minimal `tbd.out.json` file.

lib/measures/tbd/measure.rb
```diff
@@ -160,49 +231,10 @@ def run(model, runner, user_arguments)
...
-        end
-      else
-        exitTBD(runner, out_dir, nil)
-      end
-    else
-      exitTBD(runner, out_dir, nil)
-    end
 
-    return true
+    return exitTBD(runner, write_tbd_json, io, surfaces)
```
rd2, line 237: Moving most of the I/O features in _exitTBD_.

lib/log.rb
```diff
@@ -39,6 +39,14 @@ module TBD
   @@tag[ERROR] = "ERROR"
   @@tag[FATAL] = "FATAL"
 
+  @@msg = []
+  @@msg[0]     = ""
+  @@msg[DEBUG] = "For debugging"
+  @@msg[INFO]  = "No errors, no warnings"
+  @@msg[WARN]  = "Non-fatal warnings"
+  @@msg[ERROR] = "Non-fatal errors"
+  @@msg[FATAL] = "Fatal errors"
```
rd2, line 48: Temporary half-sentences for now ...

lib/measures/tbd/measure.rb
rd2: The latest commit illustrates most of what's proposed here. Seems to be working well. In a transition state as the logger branch tests still point to the master JSON schema. It still provides a good idea.

rd2: FYI, no longer in-between states (i.e., _master_ vs _logger_ branches). Fully tested against revised schema.

spec/tbd_tests_spec.rb
rd2: Schema is now enriched to log a single timestamp (everything happens under ~1 sec), log objects are kept distinct (facilitates subsequent parsing), etc.

lib/log.rb
macumber: Just a dumb question, do you have to do logging in French and English?

spec/tbd_tests_spec.rb
```diff
@@ -4869,7 +5154,267 @@
...
+      end
+      io[:logs] = msgs
+    end
+
+    out = JSON.pretty_generate(io)
+    outP = File.dirname(__FILE__) + "/../json/tbd_warehouse9.out.json"
+    File.open(outP, "w") do |outP|
+      outP.puts out
+    end
+    # ... should contain a 'logs' entry at the end of the JSON output file.
```
macumber, line 5413: For me this output just seems to show the log level, no message is included?

```
    "logs": [
        " (ERROR) ",
        " (ERROR) ",
        " (ERROR) ",
        " (ERROR) ",
        " (WARNING) ",
        " (WARNING) "
    ]
```

spec/tbd_tests_spec.rb
macumber: I agree that a timestamp for the overall conversion is just as valuable as the timestamp for each log message (as long as the messages are kept in order so you can tell which errors happen first).

tbd.schema.json
```diff
@@ -435,8 +430,41 @@
...
+          "title": "Logged messages",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/Logs"
+          }
+        }
+      },
+      "additionalProperties": false
+    },
+    "Logs": {
```
macumber, line 454: Is `Logs` still needed?

tbd.schema.json
```diff
@@ -61,17 +61,12 @@
...
-      "uniqueItems": true
+      "type": "object",
+      "$ref": "#/definitions/Building"
     },
-    "logs": {
-      "type": "array",
-      "items": {
-        "$ref": "#/definitions/Log"
-      }
+    "log": {
```
macumber, line 67: The test output I saw had `logs` instead of `log`

spec/tbd_tests_spec.rb
```diff
@@ -4869,7 +5154,267 @@
...
+    #       "id": "cantilevered beam",
+    #       "point": 0.6
+    #     }
+    #   ],
+    #   "surfaces": [
+    #     {
+    #       "id": "Office Front Wall",
+    #       "khis": [
+    #         {
+    #           "id": "beam",                           <<<< non-fatal ERROR # 2
```
macumber, line 5281: Maybe call "missing_id" for clarity?

spec/tbd_tests_spec.rb
```diff
@@ -4869,7 +5154,267 @@
...
+    #         {
+    #           "id": "cantilevered beam",
+    #           "count": 300                           <<<< triggers WARNING # 1
+    #         }
+    #       ]
+    #     }
+    #   ],
+    #   "edges": [
+    #     {
+    #       "psi": "detailed",                          <<<< non-fatal ERROR # 3
```
macumber, line 5298: Maybe call "missing_id" for clarity?

spec/tbd_tests_spec.rb
```diff
@@ -4869,7 +5154,267 @@
...
+    expect(surfaces["Office Left Wall"].has_key?(:edges)).to be(true)
+    expect(surfaces.has_key?("Fine Storage Right Wall")).to be(true)
+    expect(surfaces["Fine Storage Right Wall"].has_key?(:edges)).to be(true)
+
+    expect(TBD.status).to eq(TBD::ERROR)
+    expect(TBD.logs.size).to eq(6)
+    # TBD.logs.each { |log| puts log[:msg] }
+    #   'clerestory' vertex count (3 or 4)
+    #   Can't override 'regular (BETBG)' PSI set  - skipping
+    #   'Office Front Wall' KHI 'beam' mismatch
```
macumber, line 5328: Is this a mismatch or is this a missing key error?  Maybe "Can't find KHI named 'beam' referenced by 'Office Front Wall'" is a better error?

macumber: Do you want to tell the user what you are going to do when you skip here?  Are you going to use the default KHI or just skip derating?  Same thing for missing PSI.

json/midrise.json
```diff
@@ -1,5 +1,5 @@
 {
-  "schema": "https://github.com/rd2/tbd/blob/master/tbd.schema.json",
+  "schema": "https://github.com/rd2/tbd/blob/logger/tbd.schema.json",
```
macumber, line 2: You might want to change these back to point at master right before merging

lib/framedivider.rb
```diff
@@ -1,22 +1,22 @@
...
 #
 # @return [Array] Flattened OS 3D points array
 def flatZ(pts)
-  raise "Invalid pts (flatZ)" unless pts
-
-  cl = OpenStudio::Point3dVector
-  valid = pts.is_a?(cl) || pts.is_a?(Array)
-  raise "#{pts.class}? expected #{cl} (flatZ)" unless valid
+  unless pts && (pts.is_a?(OpenStudio::Point3dVector) || pts.is_a?(Array))
+    TBD.log(TBD::DEBUG, "Invalid pts - flatZ")
```
macumber, line 11: Is this really just a debug message?  Something has gone very wrong if you are seeing this right?

lib/conditioned.rb
```diff
@@ -221,9 +222,10 @@ def maxHeatScheduledSetpoint(zone)
   # standards/Standards.ThermalZone.rb#L910
   setpoint = nil
   dual = false
-  raise "Invalid zone (max T)" unless zone
-  cl = OpenStudio::Model::ThermalZone
-  raise "#{zone.class}? expected #{cl} (max T)" unless zone.is_a?(cl)
+  unless zone && zone.is_a?(OpenStudio::Model::ThermalZone)
+    TBD.log(TBD::DEBUG, "Invalid zone - max T")
```
macumber, line 226: The 'max T' here would confuse me if I was a user, I would wonder I do something wrong with a max T?  Really something else is going on here, it seems like you really should never get to this line in your code.  Is it really a debug level?

For your purposes you want a unique string so you can see which line of code failed.  You could add some code in your logger to get the line of code where the log statement was called using `caller`

https://stackoverflow.com/questions/6322524/how-do-you-view-a-sample-of-the-call-stack-in-ruby

lib/log.rb
rd2: Good question (not dumb). Once/if the OS App offers language options (at some point French), I'll rehaul TBD GUI options & I/O messaging/logging. But not in the short term.

spec/tbd_tests_spec.rb
rd2: Nice catch. A remnant. Fixed to better reflect what users would see (e.g., runner output, `tbd.out.json`).

tbd.schema.json
rd2: Fixed.

tbd.schema.json
rd2: Done.

spec/tbd_tests_spec.rb
rd2: More detailed in-line comments.

spec/tbd_tests_spec.rb
rd2: More informative in-line comments.

spec/tbd_tests_spec.rb
rd2: Good points. I'm revising the messages to be as informative as possible. I may have to seek a balance between minimal logged messages vs online guide material (with a section/tab on what TBD does when faced with missing, incomplete or invalid inputs),

json/midrise.json
rd2: Yup. I've had to do this in the past whenever I brought in changes to the schema. I did as you suggest - do a 2nd follow-up merge to _master_ to correct the links.

lib/conditioned.rb
rd2: I mostly agree here. First things first: as mentioned earlier, I'm currently revising the TBD log messages to be more informative (within reason) - in this case, likely reverting back to the original checks (e.g., invalid/empty _zone_ object, wrong class). Certainly more practical for 3rd-party developers relying on TBD. This addresses part of your suggestion. Otherwise, I wasn't planning to set up a _production debugging_ solution for TBD (this would require an additional _TBD Measure_ 'debug' argument, I presume). TBD DEBUG messages should not make their way in production mode. _Normal_ users (e.g., designers, modellers) would not have the option of running TBD in DEBUG mode (unless reassembling TBD), and therefore wouldn't come across these (and so shouldn't be confused). As with most errors labelled under DEBUG (in lower-level functions), TBD simply keeps track of safe _exits_ preventing Ruby crashes. For such an error to happen, something has to have gone horribly wrong upstream e.g., an invalid OSM that wasn't caught by OpenStudio (and/or the Measures code) or an internal TBD bug. If one navigates the TBD calling code for _maxHeatScheduledSetpoint_, adequate safeguards are in place to catch these. I would rather fix the upstream code then rely on lower-level functions to report invalid `.osm` file entries. If we ever felt it necessary to truly validate OSM entries beforehand, I would prefer setting up a _validOSM?_ function that would log everything wrong with an OSM from TBD/Topolys' POV - a lot of effort. Should we be parsing OpenStudio logs for such anomalies? For instance, a 5-sided window should be red-flagged by OpenStudio, IMO - is this the case?

lib/framedivider.rb
rd2: See [here](https://github.com/rd2/tbd/pull/63#discussion_r665328560). I wasn't planning on a _production debugging_ solution - more of a basic development-only usage.

lib/conditioned.rb
macumber: I think a verify OSM method would be good before doing any TDB processing, of course that can wait until you have time/resources to work on it.  My one input would be that somethings you think of as OSM errors (e.g. 5-sided window) are not necessarily going to be flagged by OpenStudio.  It probably is going to be up to TBD to verify that the OSM conforms to any TBD specific requirements.

rd2: Got it. Just to be clear for others: TBD/Topolys does not have particular requirements concerning many of these errors - EnergyPlus does. The [5-sided window](https://github.com/NREL/EnergyPlus/issues/6534) case is a good example: TBD/Topolys can handle n-sided, irregular, concave vs convex subsurfaces just fine. As it stands, OpenStudio relies on EnergyPlus to raise the 5-sided window error and halt the simulation. So be it, but it does make writing Measures more challenging as many of these errors need to raised pre-simulation. For OSM geometry/topolgy, this would not be limited to TBD/Topolys (e.g. cost estimation, AFN). So taking the `validOSM?` function (or group of functions) a step further - maybe better to extend the [OpenStudio Common Measures Gem](https://github.com/macumber/openstudio-common-measures-gem/tree/develop/lib/measures)? I'd prefer putting (future) resources on that, and have TBD either recommend users rely on a handful of those - or have TBD integrate a number of these directly (under `/resources/`). That would lighten the load for TBD and make it (more readily) available for other Measures or Ruby scripts.

PR #70, opened 2021-09-07
Fixes odd missing U-factors

rd2:
Occasionally, some surfaces wouldn't get U-factors set (if NOT derated) - Fixed.

### DIFF
--- a/lib/framedivider.rb
+++ b/lib/framedivider.rb
@@ -30,9 +30,10 @@ end
 # @param [OpenStudio::Model::Model] model An OS model
 # @param [String] id SubSurface identifier
 # @param [Array] pts Point3D array of subsurface 'id'
+# @param [Hash] surfaces TBD surface Hash (optional)
 #
 # @return [Bool] Returns true if subsurface can fit in.
-def fit?(model, id, pts)
+def fit?(model, id, pts, surfaces = nil)
   unless model && model.is_a?(OpenStudio::Model::Model)
     TBD.log(TBD::DEBUG,
       "Can't find or validate OSM (argument) for subsurface fit - skipping")
@@ -75,16 +76,50 @@ def fit?(model, id, pts)
   end
 
   parent = parent.get
+  dad = parent.nameString
+
+  if surfaces
+    unless surfaces.is_a?(Hash)
+      TBD.log(TBD::DEBUG,
+        "Can't validate TBD surfaces (argument) for fit? - skipping")
+      return 0, 0, points
+    end
+
+    unless surfaces.has_key?(parent.nameString)
+      TBD.log(TBD::DEBUG,
+        "Can't find TBD surface '#{parent.nameString}' for fit? - skipping")
+      return 0, 0, points
+    end
+  end
+
   ft = OpenStudio::Transformation::alignFace(parent.vertices).inverse
   ft_parent = (ft * parent.vertices).reverse
   ft_parent = flatZ(ft_parent)
 
   siblings = []
   model.getSubSurfaces.each do |sub|
-    next if sub.nameString == id
+    kid = sub.nameString
+    next if kid == id
     next if sub.surface.empty?
-    next unless sub.surface.get.nameString == parent.nameString
-    siblings << flatZ( (ft * sub.vertices).reverse )
+    next unless sub.surface.get.nameString == dad
+
+    if surfaces && surfaces.has_key?(dad)
+      if surfaces[dad].has_key?(:windows)        &&
+         surfaces[dad][:windows].has_key?(kid)   &&
+         surfaces[dad][:windows][kid].has_key?(:pts)
+
+         points = surfaces[dad][:windows][kid][:pts]
+         unless points.empty?
+           siblings << flatZ( (ft * points).reverse )
+         else
+           siblings << flatZ( (ft * sub.vertices).reverse )
+         end
+      else
+        siblings << flatZ( (ft * sub.vertices).reverse )
+      end
+    else
+      siblings << flatZ( (ft * sub.vertices).reverse )
+    end
   end
 
   ft_pts = flatZ(ft * pts)
@@ -144,11 +179,12 @@ end
 #
 # @param [OpenStudio::Model::Model] model An OS model
 # @param [String] id A subsurface identifier
+# @param [Hash] surfaces TBD surface Hash (optional)
 #
 # @return [Float] Returns subsurface calculated U-factor (W/m2.K)
 # @return [Float] Returns subsurface rough opening area (m2)
 # @return [Array] Returns subsurface rough opening OpenStudio 3D points
-def opening(model, id)
+def opening(model, id, surfaces = nil)
   unless model && model.is_a?(OpenStudio::Model::Model)
     TBD.log(TBD::DEBUG,
       "Can't find or validate OSM (argument) for area & vertices - skipping")
@@ -176,6 +212,20 @@ def opening(model, id)
     return 0, 0, points
   end
   dad = s.surface.get
+
+  if surfaces
+    unless surfaces.is_a?(Hash)
+      TBD.log(TBD::DEBUG,
+        "Can't validate TBD surfaces (argument) for area & vertices - skipping")
+      return 0, 0, points
+    end
+
+    unless surfaces.has_key?(dad.nameString)
+      TBD.log(TBD::DEBUG,
+        "Can't find TBD surface '#{dad.nameString}' - skipping")
+      return 0, 0, points
+    end
+  end
 
   area = s.grossArea
   if area < TOL
@@ -413,7 +463,7 @@ def opening(model, id)
   vec << OpenStudio::Point3d.new(pts[:C][:p].x, pts[:C][:p].y, pts[:C][:p].z)
   vec << OpenStudio::Point3d.new(pts[:D][:p].x, pts[:D][:p].y, pts[:D][:p].z) if four
 
-  return u, area, points unless fit?(model, id, vec)
+  return u, area, points unless fit?(model, id, vec, surfaces)
 
   tr1 = []
   tr1 << pts[:A][:p]
--- a/lib/framedivider.rb
+++ b/lib/framedivider.rb
@@ -5,148 +5,216 @@ require "openstudio"
 #
 # @param [Array] pts OpenStudio Point3D array/vector
 #
-# @return [Array] Flattened OS 3D points array
+# @return [Array] Flattened OpenStudio 3D points array
 def flatZ(pts)
   unless pts && (pts.is_a?(OpenStudio::Point3dVector) || pts.is_a?(Array))
-    TBD.log(TBD::DEBUG,
-      "Invalid pts to flatten (argument) - skipping")
-    return pts
+    TBD.log(TBD::DEBUG, "Invalid pts to flatten (argument) - skipping")
+    return OpenStudio::Point3dVector.new
   end
 
+  vec = OpenStudio::Point3dVector.new
   pts.each do |pt|
     unless pt.class == OpenStudio::Point3d
-      TBD.log(TBD::DEBUG,
-        "#{pt.class}? expected OSM 3D points to flatten - skipping")
-      next
+      msg = "#{pt.class}? expected OSM 3D points to flatten - skipping"
+      TBD.log(TBD::DEBUG, msg)
+      return OpenStudio::Point3dVector.new
     end
-    pt = OpenStudio::Point3d.new(pt.x, pt.y, 0)
+    vec << OpenStudio::Point3d.new(pt.x, pt.y, 0)
   end
-  pts
+  vec
 end
 
-##
-# Validate whether an OpenStudio subsurface can 'fit in' (vs parent, siblings)
+## Validates whether one OpenStudio polygon fits within another polygon.
 #
-# @param [OpenStudio::Model::Model] model An OS model
-# @param [String] id SubSurface identifier
-# @param [Array] pts Point3D array of subsurface 'id'
-# @param [Hash] surfaces TBD surface Hash (optional)
+# @param [Array] poly1 Point3D array of convex polygon #1
+# @param [Array] poly1 Point3D array of convex polygon #2
+# @param [String] id1 Polygon #1 identifier (optional)
+# @param [String] id2 Polygon #2 identifier (optional)
 #
-# @return [Bool] Returns true if subsurface can fit in.
-def fit?(model, id, pts, surfaces = nil)
-  unless model && model.is_a?(OpenStudio::Model::Model)
-    TBD.log(TBD::DEBUG,
-      "Can't find or validate OSM (argument) for subsurface fit - skipping")
+# @return Returns true if either polygon fits entirely within the other.
+def fits?(poly1, poly2, id1 = "", id2 = "")
+  unless poly1 && (poly1.is_a?(OpenStudio::Point3dVector) || poly1.is_a?(Array))
+    msg = ""
+    msg << "'#{id1}': " unless id1.empty?
+    msg << "Invalid polygon (fits?, 1st argument) - skipping"
+    TBD.log(TBD::DEBUG, msg)
     return false
   end
-  unless pts && (pts.is_a?(OpenStudio::Point3dVector) || pts.is_a?(Array))
-    TBD.log(TBD::DEBUG,
-      "Invalid subsurface pts to fit (argument) - skipping")
+  unless poly2 && (poly2.is_a?(OpenStudio::Point3dVector) || poly2.is_a?(Array))
+    msg = ""
+    msg << "'#{id2}': " unless id2.empty?
+    msg << "Invalid polygon (fits?, 2nd argument) - skipping"
+    TBD.log(TBD::DEBUG, msg)
+    return false
+  end
+  if poly1.empty?
+    msg = ""
+    msg << "'#{id1}': " unless id1.empty?
+    msg << "Empty polygon (fits?, 1st argument) - skipping"
+    TBD.log(TBD::DEBUG, msg)
+    return false
+  end
+  if poly2.empty?
+    msg = ""
+    msg << "'#{id2}': " unless id2.empty?
+    msg << "Empty polygon (fits?, 2nd argument) - skipping"
+    TBD.log(TBD::DEBUG, msg)
     return false
   end
 
-  pts.each do |pt|
+  poly1.each do |pt|
     unless pt.class == OpenStudio::Point3d
-      TBD.log(TBD::DEBUG,
-        "#{pt.class}? expected OSM 3D points to fit - skipping")
+      msg = "#{pt.class}? "
+      msg << "for '#{id1}': " unless id1.empty?
+      msg << "Expected OpenStudio 3D points (fits?, 1st argument - skipping"
+      TBD.log(TBD::DEBUG, msg)
+      return false
+    end
+  end
+  poly2.each do |pt|
+    unless pt.class == OpenStudio::Point3d
+      msg = "#{pt.class}? "
+      msg << "for '#{id2}': " unless id2.empty?
+      msg << "Expected OpenStudio 3D points (fits?, 2nd argument - skipping"
+      TBD.log(TBD::DEBUG, msg)
       return false
     end
   end
 
-  s = model.getSubSurfaceByName(id)
-  if s.empty?
-    TBD.log(TBD::DEBUG,
-      "Can't find '#{id}' OSM subsurface to fit - skipping")
+  ft = OpenStudio::Transformation::alignFace(poly1).inverse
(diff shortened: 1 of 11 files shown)


lib/psi.rb
```diff
@@ -3125,6 +3125,22 @@ def processTBD(
     end
   end
 
+  # Ensure deratable surfaces have U-factors (even if NOT derated).
+  surfaces.each do |id, surface|
+    next unless surface.has_key?(:deratable)
+    next unless surface[:deratable]
+    next unless surface.has_key?(:construction)
+    next if surface.has_key?(:u)
```
rd2, line 3133: Most opaque envelope surfaces get their U-factors (or `:u`) set when _thermally derating_ their targeted insulation materials. Some surfaces (e.g. those without peripheral _edges-as-thermal-bridges_) don't get derated at all (which is fine in itself) and so didn't get their U-factors set. This generates errors in UA' summaries (i.e. has no bearing on energy simulation results). Simple fix.

lib/psi.rb
```diff
@@ -1448,8 +1448,8 @@ def rsi(construction, film_RSi, temperature = 0.0)
     end
 
     # Opaque materials next.
-    unless m.to_OpaqueMaterial.empty?
-      rsi += m.to_OpaqueMaterial.get.thermalResistance
+    unless m.to_StandardOpaqueMaterial.empty?
+      rsi += m.to_StandardOpaqueMaterial.get.thermalResistance
```
rd2, line 1452: Quite the oversight. Without `standard` in there, reported RSi values got ~doubled.

lib/ua.rb
```diff
@@ -199,18 +199,16 @@ def ua_summary(surfaces, date = Time.now, version = "",
...
     next unless surface[:deratable]
     next unless surface.has_key?(:type)
     type = surface[:type]
     next unless type == :wall || type == :ceiling || type == :floor
     next unless surface.has_key?(:net)
     next unless surface[:net] > TOL
-    next unless surface.has_key?(:u) || surface.has_key?(:r)
-    next unless surface[:u] > TOL if surface.has_key?(:u)
-    next unless surface[:r] > TOL if surface.has_key?(:r)
-    surface[:u] = 1.0 / surface[:r] unless surface.has_key?(:u)
```
rd2, line 213: Another oversight. The `:r` variable strictly refers to the targeted insulation material (to derate), not the assembly as a whole.

spec/tbd_tests_spec.rb
```diff
@@ -1818,6 +1818,8 @@
       if id == "Core_ZN_ceiling"
         expect(surface[:heatloss]).to be_within(0.001).of(0)
         expect(surface.has_key?(:ratio)).to be(false)
+        expect(surface.has_key?(:u)).to be(true)
+        expect(surface[:u]).to be_within(0.001).of(0.153)
```
rd2, line 1822: Simple test (valid).

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
+# @param [Array] poly1 Point3D array of convex polygon #2
+# @param [String] id1 Polygon #1 identifier (optional)
+# @param [String] id2 Polygon #2 identifier (optional)
 #
-# @return [Bool] Returns true if subsurface can fit in.
-def fit?(model, id, pts)
-  unless model && model.is_a?(OpenStudio::Model::Model)
-    TBD.log(TBD::DEBUG,
-      "Can't find or validate OSM (argument) for subsurface fit - skipping")
+# @return Returns true if either polygon fits entirely within the other.
```
macumber, line 34: Is it allowed to share one or more edges?  What about all edges?

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
-  if s.empty?
-    TBD.log(TBD::DEBUG,
-      "Can't find '#{id}' OSM subsurface to fit - skipping")
+  ft = OpenStudio::Transformation::alignFace(poly1).inverse
+  ft_poly1 = flatZ( (ft * poly1).reverse )
+  ft_poly2 = flatZ( (ft * poly2).reverse )
+
+  ft_model = OpenStudio::Model::Model.new
+  surface1 = OpenStudio::Model::Surface.new(ft_poly1, ft_model)
+  surface2 = OpenStudio::Model::Surface.new(ft_poly2, ft_model)
```
macumber, line 90:
```suggestion
area1 = OpenStudio::getArea(ft_poly1)
area2 = OpenStudio::getArea(ft_poly2)
```

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
+
+  area1 = surface1.grossArea
+  unless area1 > TOL
```
macumber, line 93:
```suggestion
if area1.empty?
```

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
+    TBD.log(TBD::DEBUG, msg)
     return false
   end
-
-  s = s.get
-  if s.surface.empty?
-    TBD.log(TBD::DEBUG,
-      "Missing subsurface '#{id}' OSM parent surface entry - skipping")
+  area2 = surface2.grossArea
+  unless area2 > TOL
```
macumber, line 101:
```suggestion
if area2.empty?
```

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
-  parent = model.getSurfaceByName(nom)
-  if parent.empty?
-    TBD.log(TBD::DEBUG,
-      "Can't find subsurface '#{id}' OSM parent surface '#{nom}' - skipping")
+  union = OpenStudio::join(ft_poly1, ft_poly2, TOL)
+  return false if union.empty?
+  union = union.get
+  surface = OpenStudio::Model::Surface.new(union, ft_model)
+  area = surface.grossArea
+  return false if area < TOL
```
macumber, line 114:
```suggestion
area = OpenStudio::getArea(union)
return false if area.empty?
```

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
+  if area1 > area2
+    return false if area - area1 > TOL
+  else
+    return false if area - area2 > TOL
+  end
```
macumber, line 119:
```suggestion
  if area1 > area2
    return false if Math.abs(area - area1) > TOL
  else
    return false if Math.abs(area - area2) > TOL
  end
```

Probably not needed?

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
+      msg = "#{pt.class}? "
+      msg << "for '#{id2}': " unless id2.empty?
+      msg << "Expected OpenStudio 3D points (overlaps?, 2nd argument - skipping"
+      TBD.log(TBD::DEBUG, msg)
+      return false
+    end
+  end
+
+  return true if fits?(poly1, poly2)
+  return true if fits?(poly2, poly1)
```
macumber, line 181:
```suggestion
  return true if fits?(poly1, poly2)
```

`fits?` works both ways right?

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
+
+  ft = OpenStudio::Transformation::alignFace(poly1).inverse
+  ft_poly1 = flatZ( (ft * poly1).reverse )
+  ft_poly2 = flatZ( (ft * poly2).reverse )
 
-  parent = parent.get
-  ft = OpenStudio::Transformation::alignFace(parent.vertices).inverse
-  ft_parent = (ft * parent.vertices).reverse
-  ft_parent = flatZ(ft_parent)
+  ft_model = OpenStudio::Model::Model.new
```
macumber, line 187: Avoid making a new model and just use the free methods that work on arrays of vertices in https://github.com/NREL/OpenStudio/blob/develop/src/utilities/geometry/Geometry.hpp

lib/framedivider.rb
macumber, line 181:
```suggestion
```

I'd just do the checks below it's probably less efficient to do this test which might fail

lib/framedivider.rb
```diff
@@ -5,94 +5,216 @@
...
+  return false if union.empty?
+  union = union.get
+  surface = OpenStudio::Model::Surface.new(union, ft_model)
+  area = surface.grossArea
+  return false if area < TOL
+  if area1 > area2
+    return true if area - area1 > TOL
+  else
+    return true if area - area2 > TOL
   end
```
macumber, line 218:
```suggestion
```

If the result of join is not empty return true

lib/framedivider.rb
```diff
@@ -140,123 +262,33 @@ def areaHeron(pts)
...
-      TBD.log(TBD::ERROR,
-        "OSM subsurface '#{id}' U-factor unavailable - skipping")
-      return 0, 0, points
-    else
-      u = 1.0 / r
-    end
-  else
-    u = u.get
+# @return [Array] Offset Topolys 3D points (original points if failed).
+def offset(pts, width)
```
macumber, line 271: I think `OpenStudio::buffer` does this

https://github.com/NREL/OpenStudio/blob/develop/src/utilities/geometry/Intersection.hpp#L96
https://github.com/NREL/OpenStudio/blob/develop/src/utilities/geometry/Test/Intersection_GTest.cpp#L2346

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+      msg = "Subsurface '#{id}' vertex count (must be 3 or 4) - skipping"
+      TBD.log(TBD::ERROR, msg)
+      next
+    end
+
+    typ = s.subSurfaceType.downcase
+    type = :skylight
+    type = :window if typ.include?("window")
+    type = :door if typ.include?("door")
+    glazed = true if type == :door && typ.include?("glass")
```
macumber, line 526: Are windows not glazed?

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+    #    B - C - D   <<< allowed as OpenStudio/E+ subsurface?
+    #
+    four = true if s.vertices.size == 4
+
+    if s.windowPropertyFrameAndDivider.empty?
+      vec = s.vertices
+      area = gross
+    else
+      width = s.windowPropertyFrameAndDivider.get.frameWidth
+      ptz = offset(s.vertices, width)
```
macumber, line 602: I'm not sure if you would want to add `ptz` as a new wire in the Topolys Model?  It should intersect or update other wires  automatically, I guess it depends if you want to do more with it than get the area

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+    #  B
+    #
+    # Should convert (annoying) 4-point subsurface into triangle ...
+    #        A
+    #       / \
+    #      /   \
+    #     /     \
+    #    B - C - D   <<< allowed as OpenStudio/E+ subsurface?
+    #
+    four = true if s.vertices.size == 4
```
macumber, line 595:
```suggestion
    four = (s.vertices.size == 4)
```

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+
+      area2 = 0
+      if four
+        tr2 = []
+        tr2 << ptz[:A][:p]
+        tr2 << ptz[:C][:p]
+        tr2 << ptz[:D][:p]
+        area2 = areaHeron(tr2)
+      end
+      area = area1 + area2
```
macumber, line 626:
```suggestion
area = OpenStudio::getArea(vec).get
```

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+  # Test for conflicts (with fits?, overlaps?) between surfaces to determine
+  # whether to keep original points or switch to std::vector of revised
+  # coordinates, offset by Frame & Divider frame width. This will also
+  # inadvertently catch pre-existing (yet nonetheless invalid) OpenStudio inputs
+  # (without Frame & Dividers).
+  valid = true
+  subs.each do |id, sub|
+    next unless fd
+    next unless valid
+    unless fits?(surface.vertices, sub[:points])
```
macumber, line 650: Wouldn't you want to make `fits?` directional?  You don't want a base surface that fits inside the sub surface?

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+  # Option B:
+  # surface.subSurfaces.each do |s|
+
+  # Option C:
+  # model.getModelObjects.each do |s|
+    # next if s.to_SubSurface.empty?
+    # s = s.to_SubSurface.get
+
+  # Option D:
+  subsurfaces = surface.subSurfaces.sort_by { |s| s.nameString }
```
macumber, line 512: I'd prefer sorting by area so the results don't change if you rename a surface

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+  if fd
+    if valid
+      # No conflicts. Reset subsurface gross area.
+      subs.values.each { |sub| sub[:gross] = sub[:area] }
+    else
+      # One or more conflicts between the parent surface & one or more
+      # subsurfaces, or between subsurfaces. Ignore Frame & Divider offsets
+      # and revert to original vertices.
+      subs.values.each { |sub| sub[:points] = sub[:v] }
+      subs.values.each { |sub| sub[:area] = sub[:gross] }
```
macumber, line 675: Want to log this right?

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+    end
+  end
+
+  subarea = 0
+  subs.values.each { |sub| subarea += sub[:area] }
+  surf[:net] = surf[:gross] - subarea
+
+  # Tranform final Point 3D sets, and store.
+  pts = (t * surface.vertices).map { |v| Topolys::Point3D.new(v.x, v.y, v.z) }
+  surf[:points] = pts
```
macumber, line 685: Do you want to store the frame and divider points too?

lib/framedivider.rb
```diff
@@ -390,46 +422,284 @@ def opening(model, id)
...
+  surf[:minz] = ( pts.map { |p| p.z } ).min
+
+  subs.each do |id, sub|
+    pts = (t * sub[:points]).map { |v| Topolys::Point3D.new(v.x, v.y, v.z) }
+    sub[:points] = pts
```
macumber, line 690: Do you want to store the frame and divider points too if they exist?

spec/tbd_tests_spec.rb
```diff
@@ -4708,6 +4601,90 @@
...
+    vec << OpenStudio::Point3d.new(  1,  0,  2)
+    door1 = OpenStudio::Model::SubSurface.new(vec, model)
+
+    # Door1 fits?, overlaps?
+    TBD.clean!
+    expect(fits?(door1.vertices, wall.vertices)).to be(true)
+    expect(overlaps?(door1.vertices, wall.vertices)).to be(true)
+    expect(TBD.status).to eq(0)
+
+    # Order of arguments doesn't matter.
```
macumber, line 4630: I think you want the order to matter so you can test the sub surface is inside the surface but not vice versa

spec/tbd_tests_spec.rb
```diff
@@ -5206,79 +5335,113 @@
...
+    expect(surface[:windows]["w3"][:u]).to be_within(0.01).of(2.0)
+    expect(surface[:windows]["w3"].has_key?(:points)).to be(true)
+
+    ptz = surface[:windows]["w3"][:points]
+    expect(ptz.size).to eq(3)
+
+    vec = OpenStudio::Point3dVector.new
+    ptz.each { |p| vec << t * OpenStudio::Point3d.new(p.x, p.y, p.z) }
+
+    # # This window would have 2 shared edges (@right angle) with the parent.
```
macumber, line 5435: I didn't look at these tests in depth but it is very good you have them :-)

spec/tbd_tests_spec.rb
```diff
@@ -6562,9 +6768,8 @@
     outP2 = File.dirname(__FILE__) + "/../json/tbd_warehouse13.out.json"
     File.open(outP2, "w") { |outP2| outP2.puts out2 }
 
-    # Despite holding the same content, both output JSON files are not written
-    # out the same sequentially.
-    expect(FileUtils.identical?(outP, outP2)).to be(false)
+    # The JSON output file sare identical.
```
macumber, line 6771:
```suggestion
    # The JSON output files are identical.
```

lib/framedivider.rb
rd2: If ever the need arises to store them at some point, we can always revisit this. But unless there's a desire to have Topolys generate its own Radiance files, I can't see any use for them.

lib/framedivider.rb
rd2: It does - in the previous [loop](https://github.com/rd2/tbd/blob/86bde630e3331f133f55097e25017cea2e11f96f/lib/framedivider.rb#L652).

lib/framedivider.rb
rd2: Subsurfaces like skylights usually have the same area (+50 skylights in the warehouse benchmark model). I get your point, but we can at least depend on the uniqueness of the name (not the area).

lib/framedivider.rb
rd2: Of course - always. However doors may or may not de glazed (and this does matter for some energy codes - including the Canadian ones). Here, we're just tagging a TBD `:door` as glazed (or not) based on the subsurface type "Door" vs "GlassDoor".

lib/framedivider.rb
rd2: Yes. There's a series of tests [here](https://github.com/rd2/tbd/blob/86bde630e3331f133f55097e25017cea2e11f96f/spec/tbd_tests_spec.rb#L4670).

lib/framedivider.rb
rd2: Yes simpler (thanks). I also kept a `return false if area < TOL` for the moment (in case). Upcoming commit.

lib/framedivider.rb
rd2: If the Frame & Divider inputs are valid for a given surface, then those are the points that Topolys gets as input (not the original OpenStudio subsurface vertices). If there's a conflict anywhere, then Topolys gets the original vertices.

spec/tbd_tests_spec.rb
rd2: :]

lib/framedivider.rb
macumber: Ah I see your point, you can also sort by both area and name:

surface.subSurfaces.sort_by { |s| "#{s.grossArea}-#{s.nameString}" }

Either way will give you the same result on the same model (which is good).  The thing I am concerned about is getting different results when you change the model in minor ways (e.g. renaming a subsurface).  Any type of sorting is an improvement over no sorting.

spec/tbd_tests_spec.rb
```diff
@@ -4708,6 +4601,193 @@
...
+    expect(fits?(glazing.vertices, wall.vertices)).to be(true)
+    expect(overlaps?(glazing.vertices, wall.vertices)).to be(true)
+    expect(TBD.status).to eq(0)
+
+    expect(fits?(wall.vertices, glazing.vertices)).to be(true)
+    expect(overlaps?(wall.vertices, glazing.vertices)).to be(true)
+    expect(TBD.status).to eq(0)
+  end
+
+  it "can calculate offset for (frame & dividers)" do
```
macumber, line 4709: I added this test just to show how to use the `OpenStudio::buffer` method.  If you want you could update your `offset` method to use `buffer` or not, whichever you prefer.  For a non-ninety degree angle, the `buffer` method accounts for the extra height/gap/whatever it's called that is required to join the two offsets at a single point.

lib/measures/tbd/measure.rb
```diff
@@ -124,6 +124,8 @@ def exitTBD(model, runner, gen_ua = false, ref = "", setpoints = false, out = fa
 
   io[:log] = tbd_log
 
+  runner.registerInfo("out = #{out}")
```
rd2, line 127: This is temporary (debugging). Will be removed before merging.

lib/measures/tbd/measure.rb
```diff
@@ -248,7 +250,7 @@ def arguments(model = nil)
     alter_model = OpenStudio::Measure::OSArgument.makeBoolArgument("alter_model", true, false)
     alter_model.setDisplayName("Alter OpenStudio model (Apply Measures Now)")
     alter_model.setDescription("If checked under Apply Measures Now, TBD will irrevocably alter the user's OpenStudio model.")
-    alter_model.setDefaultValue(false)
+    alter_model.setDefaultValue(true)
```
rd2: This is an oversight. Very handy if `false` for _Apply Measures Now_ iterative exploration, but if left `false` in normal simulation situations, E+ results are unaffected - not a desirable outcome! So the best is to leave this to `true` by default.

PR #76, opened 2021-12-11
Next fixes

macumber:


lib/psi.rb
```diff
@@ -1850,14 +1850,12 @@ def processTBD(
     surface[:type] = :wall if typ.include?("wall")
 
     unless s.construction.empty?
-      construction = s.construction.get.to_Construction.get
+      construction = s.construction.get.to_ConstructionBase.get
```
macumber, line 1853: The issue here was that construction was an `OS:Construction:FfactorGroundFloor` object and not an `OS:Construction`, `PlanarSurface::construction` returns a `ConstructionBase`.  `ConstructionBase` doesn't have a `layers` method, I figured you didn't really need to check if `index < construction.layers.size` as long as `deratableLayer` wouldn't return an invalid layer (which it looks like it won't).

macumber: @rd2 FYI

macumber:
This fixes the two test suites so they run for me.  However, it seems like the energy results are the same for each TBD option set, is that expected?

macumber:
I'm also seeing a lot of errors with `bundle exec rake spec`

lib/psi.rb
rd2: @macumber: Thanks. Odd that this never popped up in previous testing.

PR #82, opened 2022-03-13
Uprate menu options (draft)

rd2:
First stab at measure options.

### DIFF
--- a/lib/measures/tbd/README.md
+++ b/lib/measures/tbd/README.md
@@ -19,6 +19,14 @@ ModelMeasure
 ## Arguments
 
 
+### Alter OpenStudio model (Apply Measures Now)
+For EnergyPlus simulations, leave CHECKED. For iterative exploration with Apply Measures Now, UNCHECK to preserve original OpenStudio model.
+**Name:** alter_model,
+**Type:** Boolean,
+**Units:** ,
+**Required:** true,
+**Model Dependent:** false
+
 ### Load 'tbd.json'
 Loads existing 'tbd.json' file (under '/files'), may override 'default thermal bridge' set.
 **Name:** load_tbd_json,
@@ -31,14 +39,6 @@ Loads existing 'tbd.json' file (under '/files'), may override 'default thermal b
 e.g. 'poor', 'regular', 'efficient', 'code' (may be overridden by 'tbd.json' file).
 **Name:** option,
 **Type:** Choice,
-**Units:** ,
-**Required:** true,
-**Model Dependent:** false
-
-### Alter OpenStudio model (Apply Measures Now)
-For EnergyPlus simulations, leave CHECKED. For iterative exploration with Apply Measures Now, UNCHECK to preserve original OpenStudio model.
-**Name:** alter_model,
-**Type:** Boolean,
 **Units:** ,
 **Required:** true,
 **Model Dependent:** false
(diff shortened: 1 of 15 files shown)


lib/measures/tbd/README.md
```diff
@@ -19,44 +19,116 @@ ModelMeasure
...
+### Uprate wall construction(s)
+Uprates selected wall construction(s), to meet overall Ut target
+**Name:** uprate_walls,
 **Type:** Boolean,
 **Units:** ,
-**Required:** true,
+**Required:** false,
+**Model Dependent:** false
+
+### Uprate roof construction(s)
```
macumber, line 62: I assume the `uprate_roofs` argument doesn't do anything unless `roof_ut` is set and also that `roof_ut` doesn't do anything unless `uprate_roofs` is true.  If that is the case, you might consider removing the `uprate_roofs` argument, do not uprating if  `roof_ut` is empty, and do uprating if  `roof_ut` is not empty?

macumber: Same for floors and walls

lib/measures/tbd/README.md
```diff
@@ -19,44 +19,116 @@ ModelMeasure
...
+### Floor Ut target (W/m2•K)
+Overall Ut target to meet for exposed floor construction(s)
+**Name:** floor_ut,
+**Type:** Double,
+**Units:** ,
+**Required:** false,
+**Model Dependent:** false
+
+### Wall construction(s) to 'uprate'
+Target 1x (or 'ALL') wall construction(s) to 'uprate'
```
macumber, line 103: Maybe mention here that `wall_ut` must be set?

lib/measures/tbd/measure.rb
```diff
@@ -53,122 +53,296 @@ def modeler_description
...
+        lc = s.construction.get.to_LayeredConstruction.get
+        next if walls[:c].key?(lc.nameString)
+        next if roofs[:c].key?(lc.nameString)
+        next if flors[:c].key?(lc.nameString)
+        walls[:c][lc.nameString] = {a: lc.getNetArea} if type == "wall"
+        roofs[:c][lc.nameString] = {a: lc.getNetArea} if type == "roofceiling"
+        flors[:c][lc.nameString] = {a: lc.getNetArea} if type == "floor"
+      end
+
+      walls[:c] = walls[:c].sort_by{ |k,v| v[:a] }.reverse!.to_h
```
macumber, line 123: I think you are sorting constructions by area here?  That seems like a nice touch!

macumber: And the default should be the first option?

lib/measures/tbd/README.md
rd2: Good idea. Easy fix.

rd2: Good idea to trim down the menu items. I added a (default) "NONE" string for `roof_option` (same for wall & floors).

none

I can reword the item descriptions as needed, but additions to the [Guide](https://rd2.github.io/tbd/pages/basics.html) are necessary.

lib/measures/tbd/README.md
rd2: The added "NONE" option should do the trick.

lib/measures/tbd/measure.rb
rd2: See Figure with "NONE" option:

1. "ALL roof constructions" (all-encompassing option - likely the most common option to be picked)
2. roof constructions in sequence, in order of prevalence (largest area first, smallest area last)
3. "NONE" as last (default) item

PR #93, opened 2022-10-15
Catches invalid edges (bad input)

brgix:
- Also tests Ubuntu 22.04


lib/tbd/geo.rb
```diff
@@ -250,6 +284,8 @@ def properties(model = nil, surface = nil)
     return mismatch("model", model, cl1, mth)          unless model.is_a?(cl1)
     return mismatch("surface", surface, cl2, mth)      unless surface.is_a?(cl2)
 
+    return nil unless validate(surface)
```
brgix, line 287: @macumber : Logs (and altogether skips) invalid OpenStudio surfaces - before populating Topolys. The vast majority of (valid) surfaces would nonetheless be derated by TBD - those _connected_ to invalid ones would be partially derated. Still need to apply this to subsurfaces. As is, the _validate_ method would also catch collinear vertices (< 10mm apart) along an edge - so considering deleting intermediate collinear points from the TBD/Topolys cloned surfaces. Work in progress ...

spec/tbd_tests_spec.rb
```diff
@@ -1438,211 +1438,4 @@
       expect(argh.key?(:roof_uo)).to be(false)
     end
   end
-
```
brgix: Moved over to TBD Tests [repo](https://github.com/rd2/tbd_tests).

lib/tbd/geo.rb
```diff
@@ -372,7 +372,7 @@ def properties(model = nil, surface = nil)
       log(ERR, "Skipping '#{id}': missing construction (#{mth})")    if c.empty?
       next                                                           if c.empty?
       c = c.get.to_LayeredConstruction
-      log(ERR, "Skipping '#{id}': must be a #{cl3} (#{mth})")        if c.empty?
```
brgix, line 337: OpenStudio subsurface constructions may de defined _inter alia_ as a ConstructionAirBoundary or using a WindowDataFile. For UA' calculations, TBD currently only extracts fenestration U-factors from LayeredConstructions. This may be revised in the future for WindowDataFiles. In the meantime, TBD skips over any subsurface construction type other than LayeredConstruction (while logging a WARNING - no longer an ERROR).

lib/tbd/ua.rb
```diff
@@ -235,7 +235,7 @@ def uprate(model = nil, s = {}, argh = {})
       end
 
       if coll.empty?
-        log(ERR, "No construction to uprate - skipping (#{mth})")
+        log(ERR, "No #{label} construction to uprate - skipping (#{mth})")
```
brgix, line 238: The log message is now a bit more informative on the nature of the ERROR (e.g. when a user explicitly asks to [uprate](https://rd2.github.io/tbd/pages/ut.html) exposed floors, when an OpenStudio model doesn't even hold any such exposed floor).

PR #98, opened 2022-11-25
Allows input Hash (vs JSON)

brgix:
TBD users (e.g. measure users, gem-based scripts) could only customize TBD inputs with JSON files. These changes allows TBD to process the same JSON content directly as an input Hash. Better for scripted batch runs.

### DIFF
--- a/lib/tbd/psi.rb
+++ b/lib/tbd/psi.rb
@@ -205,7 +205,7 @@ module TBD
         rimjoist:      0.300, # *
         parapet:       0.325, # *
         fenestration:  0.200, # *
-        corner:        0.300, # ** "regular (BETBG)", adj. for ext. dimensions
+        corner:        0.300, # ** not explicitely stated
         balcony:       0.500, # *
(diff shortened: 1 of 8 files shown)


lib/tbd/psi.rb
```diff
@@ -569,16 +569,21 @@ def inputs(s = {}, e = {}, argh = {})
...
+
+    pth = argh[:io_path    ]
     sch = argh[:schema_path]
 
-    if pth
-      return empty("JSON file", mth, FTL, ipt) unless File.size?(pth)
-      io = File.read(pth)
-      io = JSON.parse(io, symbolize_names: true)
-      return mismatch("io", io, Hash, mth, FTL, ipt) unless io.is_a?(Hash)
+    if pth && (pth.is_a?(String) || pth.is_a?(Hash))
```
brgix, line 578: (i) Hash or (ii) JSON file path (same Hash content) is accepted.

Issue #101, opened 2022-12-13
Party partition edges: Adiabatic vs OtherSideCoefficients


brgix:
Excerpt from the online [Guide](https://rd2.github.io/tbd/pages/basics.html):  

> And when an exposed surface holds an edge that isn’t shared by another exposed surface, it’s either a sign of geometric inconsistency (it happens), or that the building shares a demising (or “party”) partition with a neighbouring building. If the edge links an adiabatic surface, TBD tags it as a “party” thermal bridge - otherwise it falls back to “transition”.

For complete & fully-enclosed OpenStudio model geometries, the above logic is usually sound. This would equally apply well for (new) extensions to existing buildings. The use of [adiabatic](https://bigladdersoftware.com/epx/docs/22-2/input-output-reference/group-thermal-zone-description-geometry.html#buildingsurfacedetailed) _outside boundary conditions_ for demising surfaces (e.g. party walls) is suitable, but there are (better) alternatives such as _surface_ (a party wall referencing itself) or [OtherSideCoefficients](https://bigladdersoftware.com/epx/docs/22-2/input-output-reference/group-advanced-surface-concepts.html#surfacepropertyothersidecoefficients).  

---

TBD is increasingly integrated in building stock assessments with existing OpenStudio or EnergyPlus models, where the use of multipliers is not uncommon, e.g. story multipliers (on the left) vs space/zone multipliers (on the right). 

<img width="567" alt="multi" src="https://user-images.githubusercontent.com/98104985/207326696-74937e0d-bd5c-4569-aced-1e84011633e4.png">

Although recourse to multipliers is a legitimate approach to speed up simulation times, it does present TBD and Topolys with a number of challenges when identifying (un)shared edges and then correctly tagging them (or not) as thermal bridges - a previously-raised [issue](https://github.com/rd2/tbd/issues/29). For instance, a shared edge between a _multiplied_ space's interior (adiabatic) ceiling and its exterior wall will (unfortunately) be tagged as a "party" edge, as similar conditions to demising surfaces are met. We recommend (self-referencing) _surface_ (not adiabatic) _outside boundary conditions_ in such circumstances, but we do recognize that adiabatic surfaces are fairly prevalent in existing models with multipliers.

---

[...]

PR #102, opened 2022-12-13
Revises :party edge ruleset

brgix:
In direct response to the recent issue raised [here](https://github.com/rd2/tbd/issues/101).


brgix:
Validations:
- making sure [adiabatic](https://github.com/rd2/tbd_tests/blob/da4d5aa363b34055c0a890c53370d2640e73cd9f/spec/tbd_tests_spec.rb#L2654) conditions don't trigger default **party** edge tag
- making sure [OtherSideCoefficients](https://github.com/rd2/tbd_tests/blob/da4d5aa363b34055c0a890c53370d2640e73cd9f/spec/tbd_tests_spec.rb#L4741) (as outside boundary condition values) does trigger default **party** edge tag

PR #104, opened 2022-12-26
First demo of matches? change

brgix:
Allows users (or additional TBD features) to access "matches?" method with user-set tolerances (e.g. 100mm).

### DIFF
--- a/lib/measures/tbd/measure.rb
+++ b/lib/measures/tbd/measure.rb
@@ -48,6 +48,15 @@ class TBDMeasure < OpenStudio::Measure::ModelMeasure
     alter.setDefaultValue(true)
     args << alter
 
+    arg = "sub_tol"
+    dsc = "Proximity tolerance (e.g. 0.100 m) between subsurface edges, e.g. " \
+          "between near-adjacent window jambs."
+    sub_tol = OpenStudio::Measure::OSArgument.makeDoubleArgument(arg, false)
+    sub_tol.setDisplayName("Proximity tolerance (m)")
+    sub_tol.setDescription(dsc)
+    sub_tol.setDefaultValue(TBD::TOL)
+    args << sub_tol
+
     arg = "load_tbd_json"
     dsc = "Loads existing 'tbd.json' file (under '/files'), may override "     \
           "'default thermal bridge' set."
@@ -225,6 +234,7 @@ class TBDMeasure < OpenStudio::Measure::ModelMeasure
 
     argh                 = {}
     argh[:alter        ] = runner.getBoolArgumentValue("alter_model",      args)
+    argh[:sub_tol      ] = runner.getDoubleArgumentValue("sub_tol",        args)
     argh[:load_tbd     ] = runner.getBoolArgumentValue("load_tbd_json",    args)
     argh[:option       ] = runner.getStringArgumentValue("option",         args)
     argh[:write_tbd    ] = runner.getBoolArgumentValue("write_tbd_json",   args)
@@ -313,7 +323,7 @@ class TBDMeasure < OpenStudio::Measure::ModelMeasure
 
     argh[:version ]  = model.getVersion.versionIdentifier
     tbd              = TBD.process(model, argh)
-    argh[:io      ]  = tbd[:io]
+    argh[:io      ]  = tbd[:io      ]
     argh[:surfaces]  = tbd[:surfaces]
     setpoints        = TBD.heatingTemperatureSetpoints?(model)
     setpoints        = TBD.coolingTemperatureSetpoints?(model) || setpoints
--- a/lib/measures/tbd/measure.rb
+++ b/lib/measures/tbd/measure.rb
@@ -1,6 +1,6 @@
 # MIT License
 #
-# Copyright (c) 2020-2022 Denis Bourgeois & Dan Macumber
+# Copyright (c) 2020-2023 Denis Bourgeois & Dan Macumber
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -326,10 +326,6 @@ class TBDMeasure < OpenStudio::Measure::ModelMeasure
       end
 
       return false unless kva
-        # argh[:gen_kiva  ] = false if argh[:gen_kiva  ]
-        # argh[:kiva_force] = false if argh[:kiva_force]
-        # return false
-      # end
     end
 
     # Process all ground-facing surfaces as foundation-facing.
--- a/lib/measures/tbd/measure.rb
+++ b/lib/measures/tbd/measure.rb
@@ -296,6 +296,41 @@ class TBDMeasure < OpenStudio::Measure::ModelMeasure
       end
     end
 
+    # Pre-validate ground-facing constructions for KIVA.
+    if argh[:kiva_force] || argh[:gen_kiva]
+      kva = true
+
+      mdl.getSurfaces.each do |s|
+        id = s.nameString
+        construction = s.construction
+        next unless s.isGroundSurface
+
+        if construction.empty?
+          runner.registerError("Invalid construction for KIVA (#{id})")
+          kva = false if kva
+        else
+          construction = construction.get.to_LayeredConstruction
+
+          if construction.empty?
+            runner.registerError("KIVA requires layered constructions (#{id})")
+            kva = false if kva
+          else
+            construction = construction.get
+
+            unless TBD.standardOpaqueLayers?(construction)
+              runner.registerError("KIVA requires standard materials (#{id})")
+              kva = false if kva
+            end
+          end
+        end
+      end
+
+      unless kva
+        argh[:gen_kiva  ] = false if argh[:gen_kiva  ]
+        argh[:kiva_force] = false if argh[:kiva_force]
+      end
+    end
+
     # Process all ground-facing surfaces as foundation-facing.
     if argh[:kiva_force]
       argh[:gen_kiva] = true
--- a/lib/measures/tbd/measure.rb
+++ b/lib/measures/tbd/measure.rb
@@ -325,10 +325,11 @@ class TBDMeasure < OpenStudio::Measure::ModelMeasure
         end
       end
 
-      unless kva
-        argh[:gen_kiva  ] = false if argh[:gen_kiva  ]
-        argh[:kiva_force] = false if argh[:kiva_force]
-      end
+      return false unless kva
+        # argh[:gen_kiva  ] = false if argh[:gen_kiva  ]
+        # argh[:kiva_force] = false if argh[:kiva_force]
+        # return false
+      # end
     end
 
     # Process all ground-facing surfaces as foundation-facing.
--- a/lib/measures/tbd/tests/tbd_tests.rb
+++ b/lib/measures/tbd/tests/tbd_tests.rb
@@ -43,7 +43,7 @@ class TBDTest < Minitest::Test
 
     # get arguments and test that they are what we are expecting
     arguments = measure.arguments(model)
-    assert_equal(14, arguments.size)
+    assert_equal(15, arguments.size)
   end
 
   def test_no_load_tbd_json
--- a/lib/measures/tbd/tests/tbd_tests.rb
+++ b/lib/measures/tbd/tests/tbd_tests.rb
@@ -36,187 +36,217 @@ class TBDTest < Minitest::Test
   # end
 
   def test_number_of_arguments_and_argument_names
-    # create an instance of the measure
     measure = TBDMeasure.new
-    # make an empty model
     model = OpenStudio::Model::Model.new
-
-    # get arguments and test that they are what we are expecting
     arguments = measure.arguments(model)
     assert_equal(15, arguments.size)
   end
 
   def test_no_load_tbd_json
-    # create an instance of the measure
     measure = TBDMeasure.new
 
-    # Output dirs
-    seed_dir = File.join(__dir__, 'output/no_load_tbd_json/')
+    # Output directories.
+    seed_dir = File.join(__dir__, "output/no_load_tbd_json/")
     FileUtils.mkdir_p(seed_dir)
-    seed_path = File.join(seed_dir, 'in.osm')
+    seed_path = File.join(seed_dir, "in.osm")
 
-    # create runner with empty OSW
+    # Create runner with empty OSW, and example test model.
     osw = OpenStudio::WorkflowJSON.new
     osw.setSeedFile(seed_path)
     runner = OpenStudio::Measure::OSRunner.new(osw)
-
-    # create example test model
     model = OpenStudio::Model::exampleModel
     model.save(seed_path, true)
 
-    # get arguments
+    # Get measure arguments.
     arguments = measure.arguments(model)
     argument_map = OpenStudio::Measure.convertOSArgumentVectorToMap(arguments)
 
-    # Create hash of argument values. If the argument has a default that you
-    # want to use, you don't need it in the hash.
-    args_hash                   = {}
-    args_hash["option"        ] = "efficient (BETBG)"
-    args_hash["write_tbd_json"] = true
-    args_hash["gen_UA_report" ] = true
-    args_hash["wall_option"   ] = "ALL wall constructions"
-    args_hash["wall_ut"       ] = 0.5
-    # using defaults values from measure.rb for other arguments
+    # Hash of argument values (defaults from measure.rb for other arguments).
+    argh                   = {}
+    argh["option"        ] = "efficient (BETBG)"
+    argh["write_tbd_json"] = true
+    argh["gen_UA_report" ] = true
+    argh["wall_option"   ] = "ALL wall constructions"
+    argh["wall_ut"       ] = 0.5
 
-    # populate argument with specified hash value if specified
+    # Populate arguments with specified hash value if specified.
     arguments.each do |arg|
       temp_arg_var = arg.clone
-      gotit = args_hash.key?(arg.name)
-      assert(temp_arg_var.setValue(args_hash[arg.name])) if gotit
+      assert(temp_arg_var.setValue(argh[arg.name])) if argh.key?(arg.name)
       argument_map[arg.name] = temp_arg_var
     end
 
-    # run the measure
+    # Run the measure and assert that it ran correctly.
     Dir.chdir(seed_dir)
     measure.run(model, runner, argument_map)
     result = runner.result
-
-    # show the output
     show_output(result)
-
-    # assert that it ran correctly
-    assert_equal('Success', result.value.valueName)
+    assert_equal("Success", result.value.valueName)
     assert(result.warnings.empty?)
+    assert(result.errors.empty?)
 
-    # save the model to test output directory
-    #output_file_path = "#{File.dirname(__FILE__)}//output/test_output.osm"
-    #model.save(output_file_path, true)
+    # Save the model to test output directory.
+    output_path = File.join(seed_dir, "out.osm")
+    model.save(output_path, true)
   end
 
   def test_load_tbd_json
-    # create an instance of the measure
     measure = TBDMeasure.new
 
-    # Output dirs
+    # Output directories.
     seed_dir = File.join(__dir__, "output/load_tbd_json/")
     FileUtils.mkdir_p(seed_dir)
     seed_path = File.join(seed_dir, "in.osm")
 
-    # create runner with empty OSW
+    # Create runner with empty OSW, and example test model.
     osw = OpenStudio::WorkflowJSON.new
     osw.setSeedFile(seed_path)
     runner = OpenStudio::Measure::OSRunner.new(osw)
-
-    # create example test model
     model = OpenStudio::Model::exampleModel
     model.save(seed_path, true)
 
-    # copy tdb.json next to seed
+    # Copy tdb.json next to seed.
     origin_pth = File.join(__dir__, "tbd_full_PSI.json")
     target_pth = File.join(seed_dir, "tbd.json")
     FileUtils.cp(origin_pth, target_pth)
 
-    # get arguments
+    # Get measure arguments.
     arguments = measure.arguments(model)
     argument_map = OpenStudio::Measure.convertOSArgumentVectorToMap(arguments)
 
-    # Create hash of argument values. If the argument has a default that you
-    # want to use, you don't need it in the hash.
-    args_hash = {"load_tbd_json" => true}
-    # using defaults values from measure.rb for other arguments
+    # Hash of argument values (defaults from measure.rb for other arguments).
+    argh                   = {}
+    argh["load_tbd_json" ] = true
 
-    # populate argument with specified hash value if specified
+    # Populate arguments with specified hash value if specified.
     arguments.each do |arg|
       temp_arg_var = arg.clone
-      if args_hash.key?(arg.name)
-        assert(temp_arg_var.setValue(args_hash[arg.name]))
-      end
+      assert(temp_arg_var.setValue(argh[arg.name])) if argh.key?(arg.name)
       argument_map[arg.name] = temp_arg_var
     end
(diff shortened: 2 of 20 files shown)


lib/tbd/psi.rb
```diff
@@ -1771,6 +1769,73 @@ def process(model = nil, argh = {})
       end
     end
 
+    # Unless a user has set the thermal bridge type of an individual edge via
+    # JSON input, reset any subsurface's head, sill or jamb edges as (mild)
+    # transitions if in close proximity other similar subsurface edges. User-set
+    # tolerance must be greater than default TBD tolerance (0.01 m). Both edges'
+    # origin and terminal vertices must be in close proximity. Edges of unhinged
+    # subsurfaces are ignored.
```
brgix, line 1777: A more detailed explanation (with visuals) will be introduced in the User Guide, most likely a new tab dedicated to subsurfaces.

spec/tbd_tests_spec.rb
```diff
@@ -1413,8 +1419,7 @@
           sub2.each do |sb2|
             # With default tolerances, none of the subsurface edges "match" up.
             expect(TBD.matches?(sb1, sb2)).to be(false)
-
-            # Greater tolerances trigger 5x matches, as follows:
+            # Greater tolerances however trigger 5x matches, as follows:
             # "Sub Surface 7" ~ "Sub Surface 8" ~ "Sub Surface 6"
             # "Sub Surface 3" ~ "Sub Surface 5" ~ "Sub Surface 4"
             # "Sub Surface 1" ~ "Sub Surface 2"
```
brgix, line 1425: In the 'seb.osm' model:

- "Sub Surface 5" is sandwiched between "Sub Surface 3" and "Sub Surface 4"
- "Sub Surface 8" is sandwiched between "Sub Surface 7" and "Sub Surface 6"

With a user-set tolerance of 100mm, both window jambs of "Sub Surface 5" and "Sub Surface 8" would be reset as (mild) transitions (PSI factor of 0.000 W/K per m). The matching (nearby) jambs of the other subsurfaces would also be reset as (mild) transitions.

spec/tbd_tests_spec.rb
```diff
@@ -1424,6 +1429,159 @@
...
+
+      case nom
+      when "Entryway  Wall 4"     then expect(loss).to be_within(TOL).of(2.705)
+      when "Entryway  Wall 5"     then expect(loss).to be_within(TOL).of(4.820)
+      when "Entryway  Wall 6"     then expect(loss).to be_within(TOL).of(2.008)
+      when "Smalloffice 1 Wall 1" then expect(loss).to be_within(TOL).of(5.938)
+      when "Smalloffice 1 Wall 2" then expect(loss).to be_within(TOL).of(3.838)
+      when "Smalloffice 1 Wall 6" then expect(loss).to be_within(TOL).of(3.709)
+      when "Utility1 Wall 1"      then expect(loss).to be_within(TOL).of(5.472)
+      when "Utility1 Wall 5"      then expect(loss).to be_within(TOL).of(5.440)
```
brgix, line 1456: Heat loss from major thermal bridging (in W/K), _WITHOUT_ having TBD reset any window **jambs** as (mild) **transitions** as a result of a user-set proximity tolerance of e.g. 100mm.

spec/tbd_tests_spec.rb
```diff
@@ -1424,6 +1429,159 @@
...
+        expect(types.include?("jamb")).to be(true)
+        expect(types.include?("transition")).to be(true)
+        less *= (1 * 0.618) # 1x transition; 1x jamb
+      end
+
+      # 'dads[ (parent surface identifier) ]' holds TBD-estimated heat loss
+      # from major thermal bridging (in W/K per m) in the initial case. The
+      # substitution of 1x or 2x subsurface jamb edge types to (mild)
+      # transition(s) reduces the (revised) heat loss in the second case.
+      expect(loss + less).to be_within(TOL).of(dads[nom])
```
brgix, line 1584: Here, "less" corresponds to the difference in heat loss, in W/K (not W/K per m - _my bad_), from major thermal bridging before/after having TBD reset window **jambs** as (mild) **transitions** as a result of a user-set proximity tolerance of e.g. 100mm.

lib/measures/tbd/measure.rb
```diff
@@ -296,6 +296,41 @@ def run(mdl, runner, args)
...
+
+            unless TBD.standardOpaqueLayers?(construction)
+              runner.registerError("KIVA requires standard materials (#{id})")
+              kva = false if kva
+            end
+          end
+        end
+      end
+
+      unless kva
```
brgix, line 328: If massless (or otherwise invalid) materials/constructions are targeted for KIVA input generation, raise informative ERROR messages and ignore user-required KIVA processes altogether.

lib/tbd/geo.rb
```diff
@@ -610,6 +610,36 @@ def kiva(model = nil, walls = {}, floors = {}, edges = {})
...
+
+          unless standardOpaqueLayers?(construction)
+            log(ERR, "KIVA requires standard materials (see #{id})")
+            kva = false if kva
+          end
+        end
+      end
+    end
+
+    return a unless kva
```
brgix, line 641: Similar error catching as in _measure.rb_. If used as a measure, TBD error catching will prevent subsequent access to this "kiva" method here. The error catching here is required if TBD is used as a gem. Raise informative ERROR messages and ignore user-required KIVA processes altogether.

spec/tbd_tests_spec.rb
```diff
@@ -989,6 +990,53 @@
...
+    model = model.get
+
+    # Reset all ground-facing floor surfaces as "foundations".
+    model.getSurfaces.each do |s|
+      next unless s.outsideBoundaryCondition.downcase == "ground"
+      expect(s.construction.empty?).to be(false)
+      construction = s.construction.get
+      expect(s.setOutsideBoundaryCondition("Foundation")).to be(true)
+      expect(s.setConstruction(construction)).to be(true)
+    end
```
brgix, line 1010: Above is an example of a poor SDK-based KIVA setup. It is imprudent to set "Foundation" as surface boundary conditions without prior validation/adaption of KIVA-targeted layered constructions.

spec/tbd_tests_spec.rb
```diff
@@ -989,6 +990,53 @@
...
+
+    expect(io.nil?).to be(false)
+    expect(io.is_a?(Hash)).to be(true)
+    expect(io.empty?).to be(false)
+    expect(surfaces.nil?).to be(false)
+    expect(surfaces.is_a?(Hash)).to be(true)
+    expect(surfaces.size).to eq(43)
+
+    surfaces.values.each { |s| expect(s.key?(:kiva)).to be(false) }
```
brgix, line 1034: While invalid construction/material errors are indeed caught and KIVA is consequently disabled, TBD will nonetheless process thermal bridging. It is preferable/prudent to (in sequence):

- have TBD generate KIVA inputs in isolation ... i.e. with the "(non thermal bridging)" option
- have SDK users edit TBD-generated KIVA inputs ... e.g. adjust slab perimeter insulation width
- have TBD process thermal bridging for outdoor-facing opaque surfaces

lib/measures/tbd/measure.rb
```diff
@@ -325,10 +325,11 @@ def run(mdl, runner, args)
...
 
-      unless kva
-        argh[:gen_kiva  ] = false if argh[:gen_kiva  ]
-        argh[:kiva_force] = false if argh[:kiva_force]
-      end
+      return false unless kva
+        # argh[:gen_kiva  ] = false if argh[:gen_kiva  ]
+        # argh[:kiva_force] = false if argh[:kiva_force]
+        # return false
+      # end
```
brgix, line 332: EDIT (original comment: _"If massless (or otherwise invalid) materials/constructions are targeted for KIVA input generation, raise informative ERROR messages and ignore user-required KIVA processes altogether."_). After testing a number of scenarios using the measure (e.g. App, CLI), revising TBD feedback/outcome policy when faced with massless materials + KIVA (from _informative_ to _failure_).

lib/measures/tbd/tests/tbd_tests.rb
```diff
@@ -36,187 +36,217 @@ class TBDTest < Minitest::Test
...
+
+    # Get measure arguments.
+    arguments = measure.arguments(model)
+    argument_map = OpenStudio::Measure.convertOSArgumentVectorToMap(arguments)
+
+    # Hash of argument values (defaults from measure.rb for other arguments).
+    argh                   = {}
+    argh["gen_kiva_force"] = true
+
+    # POSTULATED USER ERROR : Slab on grade construction holds a massless layer.
```
brgix, line 225: No change brought to the example model. It natively holds a massless material in its slab-on-grade construction, which will trigger an error for KIVA.

spec/tbd_tests_spec.rb
```diff
@@ -990,53 +990,6 @@
     model.save(file, true)
   end
 
-  it "can invalidate KIVA inputs (smalloffice)" do
```
brgix: Moved the test to the TBD Tests [repo](https://github.com/rd2/tbd_tests/blob/b7b76534538da6241f752b26fcf06614c52cdafb/spec/tbd_tests_spec.rb#L9809).

spec/tbd_tests_spec.rb
macumber: Is there anything in Topolys that could help with this?  Like allowing a separate tolerance for subsurfaces?  Should this also apply to windows that are close to a corner (and so would share common framing)?

brgix: @macumber : Good suggestion. Given the (very) basic, constrained [rules](https://github.com/rd2/tbd/blob/6ee3d7b106f8e6f8d89df7097549baab61be2572/lib/tbd/psi.rb#L1772) under which the proposed solution currently operates, I think we're fine. If the solution were to extend to 2x _nearby_, parallel subsurface edges yet of different length (e.g. 1x door jamb in close proximity to 1x _shorter_ side window jamb), then I would see a need for Topolys to insert additional vertices along the door rough opening. Similarly, for 2x strip window jambs _nearly_ connected along a wall corner, it would be nice to have Topolys insert 2x additional vertices along the wall corner edge itself, aligned with the window jamb vertices (allowing TBD to ignore the _nearby_ shared wall edge length as a _corner_ thermal bridge). If the _nearly_ adjoined strip windows are not of the same height, then 3 or 4 vertices may need to be inserted along the wall corner edge. So this can easily snowball into something more complicated. It becomes even more challenging when the surfaces are tilted!

We shouldn't lose sight of why this feature is being added in the first place. The Figure illustrates a real case where a 3rd-party application ends up automatically splitting a single window into 2x smaller windows (as a result of automatically splitting a concave wall surface into 2x convex wall surfaces). This is a questionable outcome, but it happens. If the 2x windows actually shared vertices along their common edge, TBD would (correctly) consider the 2x subsurfaces as 1x. Without the new proximity tolerance, TBD instead treats both subsurfaces as unique and would process 4x window jambs (exaggerating heat losses from thermal bridging). The new feature allows TBD to treat the 2x split subsurfaces as one (2x window jambs instead of 4x).
ies

Long story short, the feature is there to accommodate odd (yet unfortunately common) cases like this one. In other circumstances, it provides some tolerance for legacy OpenStudio models. I'm a tad weary of overextending the feature - it shouldn't become a crutch for poor modelling techniques or questionable (automated) processes by 3rd-parties. It's up to you - I'll gladly incorporate/test any new Topolys features you may want to suggest.

lib/tbd/geo.rb
```diff
@@ -360,6 +360,7 @@ def properties(model = nil, surface = nil)
       next                                                          unless valid
       vec      = s.vertices
       area     = s.grossArea
+      mult     = s.multiplier
```
brgix, line 363: Tracking OpenStudio subsurface multipliers.

lib/tbd/psi.rb
```diff
@@ -1769,6 +1769,50 @@ def process(model = nil, argh = {})
...
+
+            # An edge may be tagged with (potentially conflicting) multipliers.
+            # This is only possible if the edge links 2 subsurfaces, e.g. a
+            # shared jamb between window & door. By default, TBD tags common
+            # subsurface edges as (mild) "transitions" (i.e. PSI 0 W/K.m), so
+            # there would be no point in assigning an edge multiplier. Users
+            # can however reset an edge type via a TBD JSON input file (e.g.
+            # "joint" instead of "transition"). It would be a very odd choice,
+            # but TBD doesn't prohibit it. If linked subsurfaces have different
+            # multipliers (e.g. 2 vs 3), TBD tracks the highest value.
```
brgix, line 1808: This will definitely makes it way to a new "subsurface" tab in the online Guide.

lib/tbd/psi.rb
```diff
@@ -1840,9 +1884,10 @@ def process(model = nil, argh = {})
     edges.each do |identifier, edge|
       next unless  edge.key?(:psi)
       rsi        = 0
-      max        = edge[:psi].values.max
-      type       = edge[:psi].key(max)
+      max        = edge[:psi   ].values.max
+      type       = edge[:psi   ].key(max)
       length     = edge[:length]
+      length    *= edge[:mult  ] if edge.key?(:mult)
```
brgix, line 1890: TBD maintains the original OpenStudio vertices of subsurfaces with multipliers, yet adjusts the lengths of each  perimeter edge (i.e. of type "fenestration", or "head", "sill" or "jamb"). A TBD user can easily compare the JSON-reported edge vertices vs edge length to determine whether multipliers were factored in.

spec/tbd_tests_spec.rb
```diff
@@ -1591,4 +1591,261 @@
...
+    model = trns.loadModel(path)
+    expect(model.empty?).to be(false)
+    model = model.get
+
+    model.getSubSurfaces.each do |sub|
+      parent    = sub.surface
+      expect(parent.empty?).to be(false)
+      parent    = parent.get
+      front_sub = parent.nameString.include?(front)
+      left_sub  = parent.nameString.include?(left)
```
brgix, line 1712: Limiting the test to the warehouse office _left_ and _front_ walls.

spec/tbd_tests_spec.rb
```diff
@@ -1591,4 +1591,261 @@
...
+          corners2[:lm] += edge[:length]
+        end
+      end
+
+      expect(  jambs2[:nb]).to eq(6) # no change vs initial, unaltered model
+      expect(  sills2[:nb]).to eq(2)
+      expect(  heads2[:nb]).to eq(3)
+      expect( grades2[:nb]).to eq(3)
+      expect(   rims2[:nb]).to eq(1)
+      expect(corners2[:nb]).to eq(1)
```
brgix, line 1789: TBD reports the same number of edges (as MAJOR thermal bridges), as the Topolys model remains consistent with the original OpenStudio model.

spec/tbd_tests_spec.rb
```diff
@@ -1591,4 +1591,261 @@
...
+        # unintended conflicts with parent and siblings (i.e. other subsurfaces
+        # sharing the same parent). As such, TBD would overestimate the total
+        # "grade" length by the added "sill" length.
+        expect( grades2[:lm]).to be_within(0.01).of(25.91 + 2 * 0.915)
+
+        # This (user-selected) discrepancy can easily be countered (by the very
+        # same user), by proportionally adjusting the selected "grade" PSI
+        # factor (using TBD JSON customization). For this reason, TBD will not
+        # raise this as an error. Nonetheless, the use of subsurface multipliers
+        # will require a clear set of recommendations in TBD's online Guide.
```
brgix, line 1821: This rationale needs to make its way to the online Guide - subsurface multipliers are tricky.

spec/tbd_tests_spec.rb
```diff
@@ -1591,4 +1591,261 @@
...
+      end
+
+      expect(surfaces[side].key?(:heatloss))
+      hloss = surfaces[side][:heatloss]
+
+      expect(hloss).to be_within(0.01).of(21.55 + extra) if side == front
+      expect(hloss).to be_within(0.01).of(10.09 + extra) if side == left
+    end
+
+    # TODO : ensure uprating & UA' remain consistent, w/w/o multipliers.
```
brgix, line 1849: TODO!

lib/tbd/ua.rb
```diff
@@ -593,77 +593,46 @@ def ua_summary(date = Time.now, argh = {})
...
+        next unless surface.key?(subs)
+
+        surface[subs].values.each do |sub|
+          next unless sub.key?(:gross)
+          next unless sub.key?(:u    )
+          next unless sub[:gross] > TOL
+          next unless sub[:u    ] > TOL
+
+          gross  = sub[:gross]
+          gross *= sub[:mult ]                               if sub.key?(:mult)
```
brgix, line 629: Keeping (better) track of subsurface multipliers. More tests (and potentially fixes) on the way.

lib/tbd/geo.rb
```diff
@@ -486,7 +486,9 @@ def properties(model = nil, surface = nil)
     end
 
     subarea = 0
-    subs.values.each { |sub| subarea += sub[:area] }
+
+    subs.values.each { |sub| subarea += sub[:area] * sub[:mult] }
```
brgix, line 490: Consistent tracking of subsurface multipliers.

lib/tbd/psi.rb
```diff
@@ -1771,13 +1769,125 @@ def process(model = nil, argh = {})
...
+    # JSON input, reset any subsurface's head, sill or jamb edges as (mild)
+    # transitions in close proximity with other similar subsurface edges.
+    # User-set tolerance must be greater than default TBD tolerance (0.01 m).
+    # Both edges' origin and terminal vertices must be in close proximity. Edges
+    # of unhinged subsurfaces are ignored.
+    if argh[:sub_tol] > TOL
+      edges.each do |id, edge|
+        nb    = 0                            # linked subsurfaces (i.e. "holes")
+        match = false
+        next if edge.key?(:io_type)          # skip if set in JSON
```
macumber, line 1826: I might be missing something but are these keys in the JSON schema, https://github.com/rd2/tbd/blob/develop/tbd.schema.json?

I see a `type` property but no `io_type`

brgix: Ultimately, they refer to the same. `io_type` holds a user-set edge [type](https://github.com/rd2/tbd/blob/420054cebfeac4cd1c17b626bbadc9d44b24e193/lib/tbd/psi.rb#L752), parsed from a TBD JSON input file, which would likely be different from the edge type TBD would normally identify from its ruleset. In the end, if `io_type` is set by the user for a given edge, TBD doesn't even try to tag the edge (see [here](https://github.com/rd2/tbd/blob/420054cebfeac4cd1c17b626bbadc9d44b24e193/lib/tbd/psi.rb#L1271)).

spec/tbd_tests_spec.rb
macumber: Thanks Denis, I think the option to split longer edges when there are nearby sub surfaces might be generally useful it doesn't seem like it would "solve" this case.  With as many crazy cases as these, it really seems like a way for the modeler to visualize the TDB JSON and make tweaks to it would be more useful than trying to automatically get everything right.

macumber: I can't see fully automating all the heat loss calculations without something near full CAD drawings including all the studs, window frames, etc.

lib/tbd/psi.rb
macumber: Do you have a PR for the guide updates?

lib/tbd/psi.rb
macumber: `length` here is just a local variable for calculations right?  It doesn't get stored back into the TBD JSON model right?  You might call it `effective_length` or something to be more clear that it isn't physical length?

spec/tbd_tests_spec.rb
```diff
@@ -1591,4 +1591,261 @@
...
+        # edges initially hold multiple thermal bridge types in memory, until a
+        # single, dominant type (based on PSI factor) is finally assigned
+        # (here, "grade" not "sill").
+        #
+        # By adding a 2x multiplier in OpenStudio, door area and perimeter have
+        # doubled while the initial subsurface vertices remain as before. This
+        # of course breaks model geometrical consistency (vs Topolys), which is
+        # expected with multipliers - they remain abstract modifiers that do not
+        # lend easily to 3D representation. It would be imprudent for
+        # TBD/Topolys to "stretch" subsurface vertex 3D position, based on
```
macumber, line 1883: Agree that you don't want to change the Topolys model.  Seems like you want to convert the quasi-physical Topolys model to a thermal model and then do all the multiplier and other adjustments on that.

lib/tbd/psi.rb
macumber: Oh I see, so the schema in the repo is only for TBD JSON input (user facing)?  I probably knew that and forgot.  It might be good to make a separate schema for the internal or output models?  Mostly just documenting it for yourself in the future.

spec/tbd_tests_spec.rb
macumber: I opened an issue in Topolys for adding vertices on edges for nearby holes.

https://github.com/automaticmagic/topolys/issues/23

brgix: - Correct: splitting a longer opaque surface edge (near 2x subsurface jamb edges) only makes sense if it's a corner. In the illustrated example, the taller surface edge splitting the 2x windows would be tagged as a (mild) _transition_ in any case (i.e. PSI 0 W/K per m) - there would be no point in splitting the taller edge into 3x sections.
- True, having users right-click/edit an edge (e.g. via SU plugin), which would require interacting with TBD/Topolys, would be great. In fact, it's an old(er) idea that's been in circulation for a while. Without that in place, users can still script their way in editing a TBD JSON object/file. Nonetheless, it remains fair to try to automate default behaviours whenever.

lib/tbd/psi.rb
brgix: Nope.

lib/tbd/psi.rb
brgix: Correct, local method variable. Doesn't feed back to JSON.

lib/tbd/psi.rb
brgix: Actually, the JSON schema is good for both input and output files. We have several examples where a TBD JSON output file is reused as a TBD JSON input file a second time around.

macumber: Do you want to add the `io_type` or other properties to the [schema](https://github.com/rd2/tbd/blob/develop/tbd.schema.json) in this PR?

lib/tbd/psi.rb
```diff
@@ -1815,40 +1815,39 @@ def process(model = nil, argh = {})
...
 
-          nb += 1
+        if holes[identifier].attributes.key?(:unhinged)
+          nb = 0 if holes[identifier].attributes[:unhinged]
+          break  if holes[identifier].attributes[:unhinged]
         end
 
-        next unless nb == 1   # only process edge if linking a single subsurface
+        nb += 1
+        match = true if nb > 1
```
brgix, line 1840: Fixing a bug from previous commits. Previous change prevented catching common _transition_ edges shared between 2x subsurfaces (regardless of `:sub_tol` value).

spec/tbd_tests_spec.rb
```diff
@@ -1916,7 +1916,153 @@
...
+    end
+
+    expect(door_grade).to eq(1)
+    expect(door_heads).to eq(1)
+    expect(door_sills).to eq(0)
+    expect(door_jambs).to eq(2)
+    expect(door_trns ).to eq(1) # 5x edges (instead of original 4x)
+
+    expect(door_jamb_m).to be_within(0.01).of(maxZ + 0.9)
+    expect(door_trns_m).to be_within(0.01).of(maxZ - 0.9) # same as sidelight
```
brgix, line 2064: Now working with bug fix.

PR #114, opened 2024-04-04
Pulls OSut v0.5.0

brgix:
Preliminary testing of OSut [v0.5.0](https://github.com/rd2/osut/pull/17) (branch "geo"), a TBD dependency. Although 99% done, there are likely minor improvements to add to OSut before releasing later this month. Very minor adjustments are expected for TBD (i.e. no anticipated change in test results).

### DIFF
--- a/lib/tbd/geo.rb
+++ b/lib/tbd/geo.rb
@@ -251,7 +251,7 @@ module TBD
   #
   # @return [Topolys::Vector3D] true normal vector of s
   # @return [nil] if invalid input (see logs)
-  def trueNormal(s = nil, r = 0)
+  def truNormal(s = nil, r = 0)
     mth = "TBD::#{__callee__}"
     cl  = OpenStudio::Model::PlanarSurface
     return mismatch("surface", s, cl, mth)   unless s.is_a?(cl)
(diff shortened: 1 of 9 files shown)


lib/tbd/geo.rb
```diff
@@ -251,7 +251,7 @@ def faces(s = {}, e = {})
   #
   # @return [Topolys::Vector3D] true normal vector of s
   # @return [nil] if invalid input (see logs)
-  def trueNormal(s = nil, r = 0)
+  def truNormal(s = nil, r = 0)
```
brgix, line 254: Distinguishes OSut's `trueNormal` from TBD's internal `truNormal`. The obvious way around this is proper namespaces, but this will do for now.

lib/tbd/psi.rb
```diff
@@ -1762,13 +1762,8 @@ def process(model = nil, argh = {})
...
 
-          angle = reference_V.angle(farthest_V)
-
-          if angle.nil?
-            invalid("#{id}: duplicate vertices (or too narrow)", mth, 0, ERR, 0)
-            angle = 0
-          end
-
+          angle  = reference_V.angle(farthest_V)
+          angle  = 0 if angle.nil?
```
brgix, line 1766: There are at least 2 circumstances that can trigger a nilled polar angle (around an edge):

 1. a "tiny" (i.e. invalid) surface, e.g. the width of one's finger
 2. when edges are "leader lines", as briefly discussed [here](https://github.com/rd2/osut/pull/17#discussion_r1530915726)

Getting the polar angle right allows TBD to distinguish concave vs convex corners, for instance. But in other cases, it has no effect (e.g. _transition_ edges). The raised error was initially justified when dealing with bad geometry (something the user should remedy), yet "leader lines" (i.e. _transition_ edges) shouldn't raise such an error. Simpler here to just mute the error entirely.

PR #115, opened 2024-04-17
Extend support to ruby 3.x

jmarrec:
Next OpenStudio SDK version will be ruby 3.2.2, so we need to be able to build against it

Similar PRs:
* https://github.com/rd2/oslg/pull/14
* https://github.com/rd2/osut/pull/19

@brgix  I'd be grateful if you could release a new version with these asap, it's going to remove some of pain of packaging it for us.

### DIFF
--- a/tbd.gemspec
+++ b/tbd.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.bindir                   = "exe"
   s.require_paths            = ["lib"]
   s.executables              = s.files.grep(%r{^exe/}) { |f| File.basename(f) }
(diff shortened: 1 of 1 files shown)


brgix:
Thanks for the head's up and PRs, @jmarrec. I'll be testing the Ruby 3x upgrade locally (TBD, OSut, OSlg + a number of projects that depend on them) before merging/releasing. No glitches == release tomorrow.

jmarrec:
Sounds like a good plan, thank you!

(You already know that, but note that this just loosens the allowable range, so it'll work the same for people on ruby 2.7.2).

Issue #119, opened 2024-11-19
KIVA : incorrect exposed perimeter calculations + resetKIVA needs fixing


brgix:
**Issue A** : When a surface references a [FoundationKiva](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_foundation_kiva.html) object, its [SurfacePropertyExposedFoundationPerimeter](https://s3.amazonaws.com/openstudio-sdk-documentation/cpp/OpenStudio-3.6.1-doc/model/html/classopenstudio_1_1model_1_1_surface_property_exposed_foundation_perimeter.html) object shall include both:

- slab/outdoor-wall edge lengths
- slab/basement-wall edge lengths

Invalid exposed foundation perimeter tallies generate severe EnergyPlus [errors](https://github.com/NREL/EnergyPlus/blob/4a17be908304e8ca2266863e76a7f790f2fcca19/src/EnergyPlus/HeatBalanceKivaManager.cc#L1037) (Line 1037). TBD has only (ever) accounted for slab/outdoor-wall edge lengths. Current RSpec tests unfortunately do not include _full_ basement cases - only _partial_ basement cases (e.g. basement with a walkout). So there was never any issue with tested perimeter tallies. Quite the oversight. Ensure full basement cases work as intended.

----

**Issue B** :  Also related to KIVA, yet unrelated to exposed foundation perimeters. TBD's `:reset_kiva` option purges existing KIVA-related objects in a model, and optionally starts over with new KIVA-related objects and settings. However, the option unintendedly forces all (other) ground-facing surfaces to inherit a "Foundation" boundary condition. That's not desirable. Fix.

----

**Issue C** : Ensure deprecated [SurfacePropertyExposedFoundationPerimeter](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.8.0-doc/model/html/classopenstudio_1_1model_1_1_surface_property_exposed_foundation_perimeter.html) `surfaceName` method is properly managed (pre-3.7.0 vs post-3.7.0). **EDIT**: Checked - no instances.

PR #120, opened 2024-11-19
First round of KIVA fixes/tests

brgix:
See [Issue](https://github.com/rd2/tbd/issues/119).

### DIFF
--- a/lib/tbd/geo.rb
+++ b/lib/tbd/geo.rb
@@ -833,8 +833,12 @@ module TBD
         next unless floors[id][:boundary].downcase == "foundation"
         next     if floors[id].key?(:kiva)
 
-        floors[id][:kiva   ] = :slab # initially slabs-on-grade
-        floors[id][:exposed] = 0.0   # slab-on-grade or walkout perimeter
+        # Initially set as slab-on-grade. Track 'exposed foundation perimeter'.
+        #   - outdoor wall/slab-on-grade edge lengths
+        #   - outdoor wall/basement slab walkout edge lengths
+        #   - basement wall/basement slab edge lengths
+        floors[id][:kiva   ] = :slab
+        floors[id][:exposed] = 0.0
 
         # Loop around current edge.
         edge[:surfaces].keys.each do |i|
--- a/lib/tbd/geo.rb
+++ b/lib/tbd/geo.rb
@@ -722,15 +722,10 @@ module TBD
       kva = true unless surface.surfacePropertyExposedFoundationPerimeter.empty?
       surface.resetAdjacentFoundation
       surface.resetSurfacePropertyExposedFoundationPerimeter
-      next unless surface.isGroundSurface
       next     if surface.outsideBoundaryCondition.capitalize == boundary
+      next unless surface.outsideBoundaryCondition.capitalize == "Foundation"
 
-      lc = surface.construction.empty? ? nil : surface.construction.get
       surface.setOutsideBoundaryCondition(boundary)
-      next if boundary == "Ground"
-      next if lc.nil?
-
-      surface.setConstruction(lc) if surface.construction.empty?
     end
 
     perimeters = model.getSurfacePropertyExposedFoundationPerimeters
@@ -834,6 +829,7 @@ module TBD
     edges.each do |code1, edge|
       edge[:surfaces].keys.each do |id|
         next unless floors.key?(id)
+
         next unless floors[id][:boundary].downcase == "foundation"
         next     if floors[id].key?(:kiva)
 
@@ -847,8 +843,9 @@ module TBD
           next unless walls[i][:boundary].downcase == "foundation"
           next     if walls[i].key?(:kiva)
 
-          floors[id][:kiva] = :basement
-          walls[i  ][:kiva] = id
+          floors[id][:kiva   ]  = :basement
+          floors[id][:exposed] += edge[:length]
+          walls[i  ][:kiva   ]  = id
         end
 
         # Loop around current edge.
@@ -862,7 +859,7 @@ module TBD
 
         # Loop around other floor edges.
         edges.each do |code2, e|
-          next if code1 == code2 #  skip - same edge
+          next if code1 == code2 # skip - same edge
 
           e[:surfaces].keys.each do |i|
             next unless i == id # good - same floor
@@ -873,8 +870,9 @@ module TBD
               next unless walls[ii][:boundary].downcase == "foundation"
               next     if walls[ii].key?(:kiva)
 
-              floors[id][:kiva] = :basement
-              walls[ii ][:kiva] = id
+              floors[id][:kiva   ]  = :basement
+              walls[ii ][:kiva   ]  = id
+              floors[id][:exposed] += e[:length]
             end
 
             e[:surfaces].keys.each do |ii|
@@ -889,6 +887,7 @@ module TBD
 
         foundation = OpenStudio::Model::FoundationKiva.new(model)
         foundation.setName("KIVA Foundation Floor #{id}")
+
         floor = model.getSurfaceByName(id)
         kiva  = false if floor.empty?
         next          if floor.empty?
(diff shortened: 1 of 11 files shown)


lib/tbd/geo.rb
```diff
@@ -722,15 +722,10 @@ def resetKIVA(model = nil, boundary = "Foundation")
       kva = true unless surface.surfacePropertyExposedFoundationPerimeter.empty?
       surface.resetAdjacentFoundation
       surface.resetSurfacePropertyExposedFoundationPerimeter
-      next unless surface.isGroundSurface
       next     if surface.outsideBoundaryCondition.capitalize == boundary
+      next unless surface.outsideBoundaryCondition.capitalize == "Foundation"
```
brgix, line 726: Boils down to either "Ground"-facing or "Foundation"-facing surfaces. If purging KIVA objects WITHOUT starting over, then "Foundation"-facing surfaces have their outside boundary condition reset to "Ground" - much safer. If instead purging KIVA objects THEN starting over, both "Ground"-facing and "Foundation"-facing surfaces are left as is. It's valid to have a model with one space having "Ground"-facing surfaces, with another having "Foundation"-facing surfaces. No longer any need to (re)set constructions.

lib/tbd/geo.rb
```diff
@@ -847,8 +843,9 @@ def kiva(model = nil, walls = {}, floors = {}, edges = {})
           next unless walls[i][:boundary].downcase == "foundation"
           next     if walls[i].key?(:kiva)
 
-          floors[id][:kiva] = :basement
-          walls[i  ][:kiva] = id
+          floors[id][:kiva   ]  = :basement
+          floors[id][:exposed] += edge[:length]
```
brgix, line 847: 1 of 2 lines where "Foundation"-facing wall/slab edge lengths get tallied as exposed foundation perimeter.

spec/tbd_tests_spec.rb
```diff
@@ -1237,6 +1239,31 @@
...
+    # Alas, both "exposures" can reset in the saved OSM. One solution is to
+    # first set the 'wall height above grade' value to 0:
+    kf = model.getFoundationKivas.first
+    expect(kf.isWallHeightAboveGradeDefaulted).to be true
+    expect(kf.wallHeightAboveGrade.round(1)).to eq(0.2)
+    expect(kf.setWallHeightAboveGrade(0)).to be true
+    expect(kf.isWallHeightAboveGradeDefaulted).to be false
+    expect(kf.wallHeightAboveGrade.round).to eq(0)
+    expect(oa15.setWindExposure("NoWind")).to be true
+    expect(oa15.setSunExposure("NoSun")).to be true
```
brgix, line 1265: By default, KIVA foundation objects have a 200mm _wall height above grade_ value, i.e. a top, 8-in section exposed to outdoor air. This seems to generate the following EnergyPlus warning:

```
** Warning ** BuildingSurface:Detailed="OPENAREA 1 WALL 5", Sun Exposure="SUNEXPOSED".
**   ~~~   ** ..This surface is not exposed to External Environment.  Sun exposure has no effect.
```

Initial attempts to get rid of the warning include resetting both wind and sun exposure AFTER setting boundary conditions to "Foundation", e.g.

```
expect(oa15.setOutsideBoundaryCondition("Foundation")).to be true
expect(oa15.setWindExposure("NoWind")).to be true
expect(oa15.setSunExposure("NoSun")).to be true
```

Alas, both "exposures" end up being reset in the saved OSM. One solution is to first set the _wall height above grade_ value to 0. Works. More tests needed (to come).

spec/tbd_tests_spec.rb
```diff
@@ -1248,7 +1275,7 @@
       if id == "Open area 1 Floor"
         expect(surface[:kiva]).to eq(:basement)
         expect(surface).to have_key(:exposed)
-        expect(surface[:exposed]).to be_within(TOL).of(8.70) # 12.6 - 3.9
+        expect(surface[:exposed]).to be_within(TOL).of(12.59) # not 8.70
```
brgix, line 1278: Exposed perimeter correctly set. Initial EnergyPlus runs (this model, plus others) without warnings or errors.

lib/tbd/psi.rb
```diff
@@ -1678,8 +1678,8 @@ def process(model = nil, argh = {})
     # around edge with respect to a reference vector (perpendicular to the
     # edge), +clockwise as one is looking in the opposite position of the edge
     # vector. For instance, a vertical edge has a reference vector pointing
-    # North - surfaces eastward of the edge are (0°,180°], while surfaces
-    # westward of the edge are (180°,360°].
+    # North - surfaces eastward of the edge are (0deg,180deg], while surfaces
+    # westward of the edge are (180deg,360deg].
```
brgix, line 1682: Unexpectedly, the _TBD Tests_ suite generates US-ASCII "\xC2" (°) encoding errors when running GitHub Actions. This seems limited to JSON files, and for Ruby 3 (i.e. SDK v3.8.0 & v3.9.0). Weird - as the very same files are processed when running _TBD_ GitHub Actions, which did not yield any encoding errors. Something of a Hail Mary to get rid of the annoyance.

Not particularly interested in keeping the ° char (instead of "deg", for instance), so either deleting or substituting all instances, in both JSON files and Ruby file comments.

spec/tbd_tests_spec.rb
```diff
@@ -1034,13 +1034,52 @@
...
+    woa17 = TBD.alignedWidth(oa17)
+    expect(woa13.round(2)).to eq(2.29)
+    expect(woa14.round(2)).to eq(2.14)
+    expect(woa15.round(2)).to eq(3.89)
+    expect(woa16.round(2)).to eq(2.45)
+    expect(woa17.round(2)).to eq(1.82)
+
+    # Assert 'exposed perimeter' of the Open Area space.
+    exp = woa13 + woa14 + woa15 + woa16 + woa17
+    expect(exp.round(2)).to eq(12.59)
```
brgix, line 1081: Regardless of SEB model changes, exposed perimeter should remain constant.

spec/tbd_tests_spec.rb
```diff
@@ -1185,119 +1224,137 @@
...
+    srfIDs = ["Open area 1 Floor"]
 
-    construction = oa15.construction.get
-    expect(oa15.setOutsideBoundaryCondition("Foundation")).to be true
-    expect(oa15.setAdjacentFoundation(foundation)).to be true
-    expect(oa15.setConstruction(construction)).to be true
+    # Incrementally change Open Area outdoor-facing walls to foundation-facing,
+    # and ensure KIVA reset works. Exposed perimeter should remain the same.
+    oaIDs.each_with_index do |oaID, i|
+      i3 = i + 3
```
brgix, line 1253: 5x outdoor-facing walls of the SEB Open Area space == 5 iterations.

spec/tbd_tests_spec.rb
```diff
@@ -1185,119 +1224,137 @@
...
+      expect(io).to be_a(Hash)
+      expect(io).to have_key(:edges)
+      expect(io[:edges].size).to eq(105 - 2 * i)
+      expect(model.foundationKivaSettings).to_not be_empty
+      expect(model.getSurfacePropertyExposedFoundationPerimeters.size).to eq(1)
+      expect(model.getFoundationKivas.size).to eq(1) # !4 ... previously purged
+
+      perimeter = model.getSurfacePropertyExposedFoundationPerimeters.first
+      expect(perimeter.totalExposedPerimeter).to_not be_empty
+      expect(perimeter.totalExposedPerimeter.get.round(2)).to eq(exp.round(2))
```
brgix, line 1303: Regardless of the number of _outdoor-facing_ vs _foundation-facing_ walls, exposed wall perimeter remains constant.

lib/tbd/psi.rb
macumber: Weird that characters in comments caused an issue, but this must trigger some file encoding, possibly when checked out by GitHub?

brgix: Good point. I suspect that the only culprit is the _tbd.schema.json_ file. Just decided to purge/replace ° in ALL files, regardless.

